PR #29, opened 2019-07-02
Allow subscribing to Kubernetes events by concrete object name

zuzzas:
#### The Problem
1. There are some object that lack labels. For example: Service and corresponding Endpoint `kubernetes` in `default` Namespace. I may have an interest in this exact object (whose name is the same across all Kubernetes clusters on planet Earth), but I don't want to muck around with labelings.
2. If I were to use [resourceNames](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.15/#policyrule-v1-rbac-authorization-k8s-io) field in Role or ClusterRole, I'd find that client-go's Reflector component would fail with an error:
```
E0628 13:53:29.852940       1 reflector.go:125] github.com/flant/shell-operator/pkg/kube_events_manager/kube_events_manager.go:350: Failed to list *unstructured.Unstructured: endpoints is forbidden: User "system:serviceaccount:kube-system:node-local-dns" cannot list resource "endpoints" in API group "" in the namespace "kube-system"
```
#### The Solution
Specify a `FieldSelector` that matches on exact `.metadata.name`. It works great in conjunction with `resourceNames` field and allows for a much more fine-grained event subscriptions.

### DIFF
--- a/HOOKS.md
+++ b/HOOKS.md
@@ -107,6 +107,7 @@ Syntax:
           ...
         ],
       },
+      "objectName": "backend-7zxyf",
       "namespaceSelector": {
         "matchNames": ["somenamespace", "proj-production", "proj-stage"],
         "any": true|false,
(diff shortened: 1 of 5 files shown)


zuzzas:
Unfortunately, I haven't been able to accept a list of `objectName`s since Kubernetes *field selectors* do not support OR operations, only AND.

diafour:
Dobrze zrobiony!
Events from a list of `objectName`s can be implemented with multiple informers, as it done for namespaceSelector.matchNames. Anyway, lgtm.

PR #55, opened 2019-10-01
fix: watchEvent should be array

### LINKED ISSUE #54: monitor-pods example stopped working after merging #52
flant/shell-operator:latest failed to start up with monitor-pods example hook over the weekend

```
2019-09-30T09:25:35Z ERROR : MAIN Fatal: initialize hook manager: creating hook 'pods-hooks.sh': load hook 'pods-hooks.sh' config: 2 errors occurred:
 * kubernetes.watchEvent must be of type string: "array"
 * kubernetes.watchEvent should be one of [Added Modified Deleted]
hook --config output: {
  "configVersion":"v1",
  "kubernetes":[{
    "apiVersion": "v1",
    "kind": "Pod",
    "watchEvent":["Added"]
  }]
}
```

Reverting to flant/shell-operator:v1.0.0-beta.5 (3 month ago) works again for the same hooks

diafour:
fix #54

### DIFF
--- a/pkg/hook/config/schemas.go
+++ b/pkg/hook/config/schemas.go
@@ -109,11 +109,15 @@ properties:
         kind:
           type: string
         watchEvent:
-          type: string
-          enum:
-          - Added
-          - Modified
-          - Deleted
+          type: array
+          additionalItems: false
+          minItems: 1
+          items:
+            type: string
+            enum:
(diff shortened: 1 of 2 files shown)


diafour:
@distol I think i can merge this fix without your review.

PR #58, opened 2019-10-07
CHANGE link to cron package documentation

azalio:
Old link shows us a new version of package documentation, but you use v2 version, not v3. It can mislead readers.


diafour:
:+1:

PR #108, opened 2020-01-30
feat: generate binding contexts for schedule bindings

nabokihms:
### Description:
The PR adds the ability to generate binding contexts for schedules. 

### Motivation:
After adding includeSnapshotsFrom option to schedule config we need to generate binding contexts for it too to cover hooks, that take Kubernetes cluster state periodically.

### DIFF
--- a/test/hook/context/generator.go
+++ b/test/hook/context/generator.go
@@ -17,7 +17,8 @@ import (
 	. "github.com/flant/shell-operator/pkg/hook/binding_context"
 	"github.com/flant/shell-operator/pkg/hook/controller"
 	"github.com/flant/shell-operator/pkg/kube"
-	manager "github.com/flant/shell-operator/pkg/kube_events_manager"
+	bindingmanager "github.com/flant/shell-operator/pkg/kube_events_manager"
+	schedulemanager "github.com/flant/shell-operator/pkg/schedule_manager"
 )
 
 var KubeClient kube.KubernetesClient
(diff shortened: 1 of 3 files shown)


test/hook/context/generator.go
```diff
@@ -34,14 +35,15 @@ func convertBindingContexts(bindingContexts []BindingContext) (string, error) {
...
-	Controller   StateController
-	Manager      manager.KubeEventsManager
-	Context      context.Context
-	Cancel       context.CancelFunc
+	HookCtrl        controller.HookController
+	HookMap         map[string]string
+	HookConfig      string
+	InitialState    string
+	Controller      StateController
+	BindingManager  bindingmanager.KubeEventsManager
```
diafour: Why bingingmanager? `schedule` is also a bingind ;)

https://github.com/flant/shell-operator/blob/master/pkg/hook/hook_manager.go#L41-L42

test/hook/context/generator.go
```diff
@@ -109,12 +113,13 @@ func (b *BindingContextController) Run() (string, error) {
 	}
 
 	b.HookCtrl = controller.NewHookController()
-	b.HookCtrl.InitKubernetesBindings(testHook.GetConfig().OnKubernetesEvents, b.Manager)
+	b.HookCtrl.InitKubernetesBindings(testHook.GetConfig().OnKubernetesEvents, b.BindingManager)
+	b.HookCtrl.InitScheduleBindings(testHook.GetConfig().Schedules, b.ScheduleManager)
+	b.HookCtrl.EnableScheduleBindings()
```
diafour: This call to EnableScheduleBindings() is redundant. There is no need of fully initialized ScheduleManager object because BindingContext is generated on demand by RunSchedule and not by channel monitoring.

Back then ScheduleManager was used to return error on initialization if `crontab` was not valid. Now, this check is performed when the configuration is parsed:

```
110: testHook, err = testHook.WithConfig([]byte(b.HookConfig))
```

I see no reason to start full-fledged ScheduleManager, it can be completely removed.

`nil` is good for previous line:
```
b.HookCtrl.InitScheduleBindings(testHook.GetConfig().Schedules, nil)
```

diafour: Wrong: b.HookCtrl.EnableScheduleBindings() is necessary.

PR #114, opened 2020-02-05
feat: kubeclient: methods to get APIResource objects from cluster

diafour:
- remove GroupVersionResourceByKind


diafour:
- GroupVersionResource is used to address a resource by Dynamic client
- ApiResource can be used to get a `namespaced` flag to properly list/get/delete/update resources with Dynamic client

PR #119, opened 2020-02-10
ci: build dev image from branch when pull request is labeled

diafour:


diafour:
Let's merge and see what's happening :smiling_imp:

PR #133, opened 2020-03-09
Documentation proofreading

zuzzas:
TODO: 

~- [ ] Replace every JSON example with YAML~

There should be a way to properly present available syntax options without resorting to EBNF notation or mocking a JSON document...


diafour:
https://github.com/flant/shell-operator/issues/137

PR #173, opened 2020-05-11
Fix example pods-hook.sh YAML

nhinds:
Remove stray `"` from the README example to avoid the following error:
> MAIN Fatal: initialize hook manager: creating hook 'pods-hook.sh': load hook 'pods-hook.sh' config: 1 error occurred:\n\t* .kubernetes\" is a forbidden property\n\n\nhook --config output: configVersion: v1\nkubernetes\":\n- apiVersion: v1\n  kind: Pod\n  executeHookOnEvent: [\"Added\"]\n\n


diafour:
Tnak you!

PR #195, opened 2020-08-14
issue-192: iterate through events array in namespace example

vasili439:
Fix for bash handler script in order to iterate through BINDING_CONTEXT_PATH array


diafour:
Thank you.

PR #220, opened 2020-12-07
feat: Warn users about non-executable hooks

nabokihms:
This PR is aimed to improve the UX part. Without logs, it is confusing that hooks were skipped for some reason.

### DIFF
--- a/pkg/utils/file/file.go
+++ b/pkg/utils/file/file.go
@@ -57,7 +57,7 @@ func RecursiveGetExecutablePaths(dir string) ([]string, error) {
 		}
 
 		if !IsFileExecutable(f) {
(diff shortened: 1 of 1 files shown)


pkg/utils/file/file.go
```diff
@@ -54,10 +56,12 @@ func RecursiveGetExecutablePaths(dir string) ([]string, error) {
 			return nil
 		}
 
-		if IsFileExecutable(f) {
-			paths = append(paths, path)
+		if !IsFileExecutable(f) {
+			log.Infof("file '%s' in a hooks dir skipped because it has no executable permissions", path)
```
diafour, line 60:
```suggestion
			log.Warnf("File '%s' is skipped: no executable permissions, chmod +x is required to run this hook", path)
```

PR #224, opened 2020-12-09
Kubernetes object manipulation framework proposal

zuzzas:
An attempt to document upcoming implementation.

### DIFF
--- a/fep/000-manipulating-kubernetes-objects.md
+++ b/fep/000-manipulating-kubernetes-objects.md
@@ -1,0 +1,132 @@
+# Manipulating Kubernetes objects via an opinionated CRUD framework
+
+## Summary
+
+### Goals
+
+Provide an opinionated CRUD framework for Kubernetes' object manipulation.
+
+### Non-goals
+
+Provide direct access to Kubernetes' REST API.
+
+## Manipulations
+
+Let us start with enumerating operationTypes that this framework shall support.
+These will implement all common object manipulation patterns.
+
+### Creating
+
+* `CreateorUpdate` — accept a Kubernetes object.
+  It retrieves an object, and if it already exists computes a JSON Merge Patch and applies it (will not update .status field).
+  If it does not exist, we create the object.
+
+Optionally, we might support the following Manipulations:
+
+* `Create` — will fail if an object already exists
+* `CreateIfNotExists` — create and object if such an object does not already 
+  exist by namespace/name.
+
+These manipulations are potentially dangerous and useless, since they do not reconcile an object in any way.
+
+### Deleting
+
+Mirrors all REST DELETE [options](https://kubernetes.io/docs/concepts/workloads/controllers/garbage-collection).
+
+* `EnsureDeleted`
+* `EnsureDeleteBackground` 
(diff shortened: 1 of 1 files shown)


fep/000-manipulating-kubernetes-objects.md
```diff
@@ -0,0 +1,132 @@
...
+
+##### Fire and forget
+
+These work similar to their Optimistic-locking counterparts with one crucial difference. If there is a `resourceVersion`
+conflict, shell-operator shall not call the hook again. Instead, it will apply jq filter or transformations in the anonymous Go function,
+until resource update is successful. With [an exponential backoff](https://pkg.go.dev/k8s.io/client-go@v0.19.4/util/retry#RetryOnConflict)
+and upper limit, of course.
+
+* `JQPatch`
+* `FilterPatch`
```
nabokihms, line 71: As I see it, JQPatch is a special case of a FilterPatch. We will generate an anonymous function based on a jq-expression and use it internally. Hooks that have written in Go will be able to set this anonymous function directly. 

I am a little bit puzzled because internal concepts and user interface mixed :)

nabokihms (commented):
Think we need to discuss it internally before proceeding.

fep/000-manipulating-kubernetes-objects.md
```diff
@@ -0,0 +1,132 @@
...
+will describe desired operation.
+
+### Path
+
+Environment variable `$KUBERNETES_OPERATION_PATH`.
+
+### JSON format
+
+* `operationType` — specifies an operation's type.
+* `apiVersion` — optional field that specifies object's apiVersion. If not present, we'll deduce an object's preferred
```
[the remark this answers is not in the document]
zuzzas: It is not. If a user would like to be sure, that he operates on the object from this exact version, he should be able to do so. If they do not care about this, for example, while changing `nodeSelector` (a field that is unlikely to ever be changed), then they skip `apiVersion` field.

fep/000-manipulating-kubernetes-objects.md
```diff
@@ -0,0 +1,132 @@
...
+  apiVersion from `discovery` client.
+* `kind` — object's Kind.
+* `namespace` — object's Namespace. If empty, implies operation on a Cluster-level resource.
+* `statusPatch` — optional field. If true, when modifying an object, send requests to `/status` subresource.
```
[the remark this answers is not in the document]
zuzzas: It does. Unfortunately, semantics change heavily once we deviate from `/` and `/status`.

For example, if we were to use the `/scale` subresource, it does not expect a resource object, but a `v1.Scale` object. Or a patch to it.

fep/000-manipulating-kubernetes-objects.md
zuzzas: That is true. Let's see how I can rephrase this...

diafour: Go hooks and shell hooks are separate right now.

zuzzas:
We'll skip the proposal part on such small changes. :)

PR #227, opened 2020-12-10
feat: metrics with hook resource usage

diafour:
- cpu sys and user milliseconds
- max rss kilobytes

### DIFF
--- a/pkg/shell-operator/metrics.go
+++ b/pkg/shell-operator/metrics.go
@@ -97,6 +97,35 @@ func RegisterHookMetrics(metricStorage *metric_storage.MetricStorage) {
 		},
 	)
 
+	// System CPU usage.
+	metricStorage.RegisterHistogramWithBuckets(
+		"{PREFIX}hook_run_user_cpu_seconds",
+		labels,
+		[]float64{
+			0.0,
+			0.001, 0.002, 0.005, // 1,2,5 milliseconds
+			0.01, 0.02, 0.05, // 10,20,50 milliseconds
+			0.1, 0.2, 0.5, // 100,200,500 milliseconds
+			1, 2, 5, // 1,2,5 seconds
+			10, // 10 seconds
+		},
+	)
+	// User CPU usage.
+	metricStorage.RegisterHistogramWithBuckets(
+		"{PREFIX}hook_run_sys_cpu_seconds",
+		labels,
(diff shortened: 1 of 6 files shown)


zuzzas (changes requested):
We've decided to replace counters with histograms.

pkg/shell-operator/operator.go
```diff
@@ -428,7 +428,14 @@ func (op *ShellOperator) TaskHandleHookRun(t task.Task) queue.TaskResult {
 		}
 	}
 
-	metrics, err := taskHook.Run(hookMeta.BindingType, hookMeta.BindingContext, hookLogLabels)
+	usage, metrics, err := taskHook.Run(hookMeta.BindingType, hookMeta.BindingContext, hookLogLabels)
+
+	if usage != nil {
+		log.Infof("Usage: %+v", usage)
+		op.MetricStorage.CounterAdd("{PREFIX}hook_run_usage_sys_ms_total", float64(usage.Sys.Milliseconds()), metricLabels)
```
nabokihms, line 435: Perhaps, it will be better to convert metrics values to seconds and bytes.
Most metrics use these units in various exporters.

Although we can leave it as is, but I would like to replace short names with full (ms -> milliseconds, kb -> kilobytes).

diafour: Yes, seconds and bytes are better.

zuzzas (approved):

pkg/shell-operator/metrics.go
```diff
@@ -97,6 +97,35 @@ func RegisterHookMetrics(metricStorage *metric_storage.MetricStorage) {
...
+		[]float64{
+			0.0,
+			0.001, 0.002, 0.005, // 1,2,5 milliseconds
+			0.01, 0.02, 0.05, // 10,20,50 milliseconds
+			0.1, 0.2, 0.5, // 100,200,500 milliseconds
+			1, 2, 5, // 1,2,5 seconds
+			10, // 10 seconds
+		},
+	)
+	// Max RSS in kilobytes.
```
nabokihms, line 126:
```suggestion
	// Max RSS in bytes.
```

diafour: Done

nabokihms (approved):
lgtm

PR #239, opened 2020-12-25
Object manipulation framework

zuzzas:
#### Overview

Provide a way to delegate Kubernetes object manipulation to the shell-operator. Will remove the need to use kubectl.

```release-note

```

### DIFF
--- a/KUBERNETES.md
+++ b/KUBERNETES.md
@@ -117,7 +117,7 @@ rapidly changing object, for example `status` field with many concurrent changes
 * `kind` — object's Kind.
 * `namespace` — object's Namespace. If empty, implies operation on a Cluster-level resource.
 * `name` — object's name.
-* `jqExpression` — describes transformations to perform on an object.
+* `jqFilter` — describes transformations to perform on an object.
 * `subresource` — e.g., `status`.
 
 ##### Example
@@ -128,7 +128,7 @@ rapidly changing object, for example `status` field with many concurrent changes
   "kind": "Deployment",
   "namespace": "default",
   "name": "nginx",
-  "jqExpression": ".spec.replicas = 1"
+  "jqFilter": ".spec.replicas = 1"
 }
 ```
 
@@ -147,7 +147,7 @@ rapidly changing object, for example `status` field with many concurrent changes
 
 ```json
 {
-  "operationType": "MergePatch",
+  "operation": "MergePatch",
   "kind": "Deployment",
   "namespace": "default",
   "name": "nginx",
@@ -167,14 +167,14 @@ rapidly changing object, for example `status` field with many concurrent changes
 * `kind` — object's Kind.
 * `namespace` — object's Namespace. If empty, implies operation on a Cluster-level resource.
 * `name` — object's name.
-* `jsonPatch` — describes transformations to perform on an object.
+* `JSONPatch` — describes transformations to perform on an object.
 * `subresource` — e.g., `status`.
 
 ##### Example
 
 ```json
 {
-  "operationType": "JSONPatch",
+  "operation": "JSONPatch",
   "kind": "Deployment",
   "namespace": "default",
   "name": "nginx",
--- a/KUBERNETES.md
+++ b/KUBERNETES.md
@@ -91,7 +91,7 @@ The path to the file is found in the `$KUBERNETES_PATCH_PATH` environment variab
 * `kind` — object's Kind.
 * `namespace` — object's namespace. If empty, implies operation on a cluster-level resource.
 * `name` — object's name.
-* `subresource` — e.g., `status`.
+* `subresource` — a subresource name if subresource is to be transformed. For example, `status`.
 
 #### Example
 
@@ -109,6 +109,10 @@ The path to the file is found in the `$KUBERNETES_PATCH_PATH` environment variab
 Use `JQPatch` for almost everything. Consider using `MergePatch` or `JSONPatch` if you are attempting to modify 
 rapidly changing object, for example `status` field with many concurrent changes (and incrementing `resourceVersion`).
 
+Be careful, when updating a `.status` field. If a `/status` subresource is enabled on a resource,
+it'll ignore updates to the `.status` field if you haven't specified `subresource: status` in the operation spec.
+More info [here](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#spec-and-status).
+
 #### JQPatch
 
 * `operation` — specifies an operation's type.
@@ -118,8 +122,7 @@ rapidly changing object, for example `status` field with many concurrent changes
 * `namespace` — object's Namespace. If empty, implies operation on a Cluster-level resource.
 * `name` — object's name.
 * `jqFilter` — describes transformations to perform on an object.
-* `subresource` — e.g., `status`.
-
+* `subresource` — a subresource name if subresource is to be transformed. For example, `status`.
 ##### Example
 
 ```json
@@ -167,8 +170,8 @@ rapidly changing object, for example `status` field with many concurrent changes
 * `kind` — object's Kind.
 * `namespace` — object's Namespace. If empty, implies operation on a Cluster-level resource.
 * `name` — object's name.
-* `JSONPatch` — describes transformations to perform on an object.
-* `subresource` — e.g., `status`.
+* `jsonPatch` — describes transformations to perform on an object.
+* `subresource` — a subresource name if subresource is to be transformed. For example, `status`.
 
 ##### Example
 
--- a/KUBERNETES.md
+++ b/KUBERNETES.md
@@ -7,8 +7,6 @@ List of possible operations and corresponding JSON specifications can be found b
 
 The path to the file is found in the `$KUBERNETES_PATCH_PATH` environment variable.
 
-> TODO: bash examples
-
 ## Operations
 
 ### Create
--- a/KUBERNETES.md
+++ b/KUBERNETES.md
@@ -1,0 +1,185 @@
+# Modifying Kubernetes objects
+
+You can delegate Kubernetes object manipulation to the shell-operator.
+
+To do this, you have to write one or more JSON or YAML documents describing operation type and its parameters to a file.
+List of possible operations and corresponding JSON specifications can be found below.
+
+The path to the file is found in the `$KUBERNETES_PATCH_PATH` environment variable.
+
+> TODO: bash examples
+
+## Operations
+
+### Create
+
+* `operation` — specifies an operation's type.
+    * `CreateOrUpdate` — accept a Kubernetes object.
+      It retrieves an object, and if it already exists, computes a JSON Merge Patch and applies it (will not update .status field).
+      If it does not exist, we create the object.
+    * `Create` — will fail if an object already exists
+    * `CreateIfNotExists` — create an object if such an object does not already
+      exist by namespace/name.
+* `object` — full object specification including "apiVersion", "kind" and all necessary metadata.
+
+#### Example
+
+```json
+{
+  "operation": "CreateOrUpdate",
+  "object": {
+    "apiVersion": "apps/v1",
+    "kind": "DaemonSet",
+    "metadata": {
+      "name": "flannel",
+      "namespace": "d8-flannel"
+    },
+    "spec": {
+      "selector": {
+        "matchLabels": {
+          "app": "flannel"
+        }
+      },
+      "template": {
+        "metadata": {
+          "labels": {
+            "app": "flannel",
+            "tier": "node"
+          }
+        },
+        "spec": {
+          "containers": [
+            {
+              "args": [
+                "--ip-masq",
+                "--kube-subnet-mgr"
+              ],
+              "image": "flannel:v0.11",
+              "name": "kube-flannel",
+              "securityContext": {
+                "privileged": true
+              }
+            }
+          ],
+          "hostNetwork": true,
+          "imagePullSecrets": [
+            {
+              "name": "registry"
+            }
+          ],
+          "terminationGracePeriodSeconds": 5
+        }
+      },
+      "updateStrategy": {
+        "type": "RollingUpdate"
+      }
+    }
+  }
+}
+```
+
+### Delete
+
+* `operation` — specifies an operation's type. Deletion types map directly to Kubernetes
+  DELETE's [`propagationPolicy`](https://kubernetes.io/docs/concepts/workloads/controllers/garbage-collection/).
+  * `Delete` — foreground deletion. Hook will block the queue until the referenced object and all its descendants are deleted.
+  * `DeleteInBackground` — will delete the referenced object immediately. All its descendants will be removed by Kubernetes'
+    garbage collector.
+  * `DeleteNonCascading` — will delete the referenced object immediately, and orphan all its descendants.
+* `apiVersion` — optional field that specifies object's apiVersion. If not present, we'll use preferred apiVersion
+  for the given kind.
+* `kind` — object's Kind.
+* `namespace` — object's namespace. If empty, implies operation on a cluster-level resource.
+* `name` — object's name.
+* `subresource` — e.g., `status`.
+
+#### Example
+
+```json
+{
+  "operation": "Delete",
+  "kind": "Pod",
+  "namespace": "default",
+  "name": "nginx"
+}
+```
+
+### Patch
+
+Use `JQPatch` for almost everything. Consider using `MergePatch` or `JSONPatch` if you are attempting to modify 
+rapidly changing object, for example `status` field with many concurrent changes (and incrementing `resourceVersion`).
+
+#### JQPatch
+
+* `operation` — specifies an operation's type.
+* `apiVersion` — optional field that specifies object's apiVersion. If not present, we'll use preferred apiVersion
+  for the given kind.
+* `kind` — object's Kind.
+* `namespace` — object's Namespace. If empty, implies operation on a Cluster-level resource.
+* `name` — object's name.
+* `jqExpression` — describes transformations to perform on an object.
+* `subresource` — e.g., `status`.
+
+##### Example
+
+```json
+{
+  "operation": "JQPatch",
+  "kind": "Deployment",
+  "namespace": "default",
+  "name": "nginx",
+  "jqExpression": ".spec.replicas = 1"
+}
+```
+
+#### MergePatch
+
+* `operation` — specifies an operation's type.
+* `apiVersion` — optional field that specifies object's apiVersion. If not present, we'll use preferred apiVersion
+  for the given kind.
+* `kind` — object's Kind.
+* `namespace` — object's Namespace. If empty, implies operation on a Cluster-level resource.
+* `name` — object's name.
+* `mergePatch` — describes transformations to perform on an object.
+* `subresource` — e.g., `status`.
+
+##### Example
+
+```json
+{
+  "operationType": "MergePatch",
+  "kind": "Deployment",
+  "namespace": "default",
+  "name": "nginx",
+  "mergePatch": {
+    "spec": {
+      "replicas": 1
+    }
+  }
+}
+```
+
+#### JSONPatch
+
+* `operation` — specifies an operation's type.
+* `apiVersion` — optional field that specifies object's apiVersion. If not present, we'll use preferred apiVersion
+  for the given kind.
+* `kind` — object's Kind.
+* `namespace` — object's Namespace. If empty, implies operation on a Cluster-level resource.
+* `name` — object's name.
+* `jsonPatch` — describes transformations to perform on an object.
+* `subresource` — e.g., `status`.
+
+##### Example
+
+```json
+{
+  "operationType": "JSONPatch",
+  "kind": "Deployment",
+  "namespace": "default",
+  "name": "nginx",
+  "jsonPatch": [
+    {"op": "replace", "path":  "/spec/replicas", "value":  1}
+  ]
+}
+```
--- a/go.mod
+++ b/go.mod
@@ -21,6 +21,7 @@ require (
 	gopkg.in/alecthomas/kingpin.v2 v2.2.6
 	gopkg.in/robfig/cron.v2 v2.0.0-20150107220207-be2e0b0deed5
 	gopkg.in/satori/go.uuid.v1 v1.2.0
+	gopkg.in/yaml.v3 v3.0.0-20191120175047-4206685974f2
 	k8s.io/api v0.17.0
 	k8s.io/apiextensions-apiserver v0.17.0
 	k8s.io/apimachinery v0.17.0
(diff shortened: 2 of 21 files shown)


KUBERNETES.md
```diff
@@ -0,0 +1,134 @@
...
+
+The path to the file is found in the `$KUBERNETES_OPERATIONS_PATH` environment variable.
+
+> TODO: bash examples
+
+## Operations
+
+### Create
+
+* `operationType` — specifies an operation's type. Deletion types map directly to Kubernetes
```
diafour, line 16: Maybe just "operation"? Also, deletion info seems redundant here.

KUBERNETES.md
```diff
@@ -0,0 +1,134 @@
...
+    * `CreateIfNotExists` — create an object if such an object does not already
+      exist by namespace/name.
+* `object` — full object specification including "apiVersion", "kind" and all necessary metadata.
+
+#### Example
+
+```json
+{
+  "operationType": "CreateOrUpdate",
+  "object": "{\"apiVersion\":\"apps/v1\",\"kind\":\"DaemonSet\",\"metadata\":{\"name\":\"flannel\",\"namespace\":\"d8-flannel\"},\"spec\":{\"selector\":{\"matchLabels\":{\"app\":\"flannel\"}},\"template\":{\"metadata\":{\"labels\":{\"app\":\"flannel\",\"tier\":\"node\"}},\"spec\":{\"containers\":[{\"args\":[\"--ip-masq\",\"--kube-subnet-mgr\"],\"image\":\"flannel:v0.11\",\"name\":\"kube-flannel\",\"securityContext\":{\"privileged\":true}}],\"hostNetwork\":true,\"imagePullSecrets\":[{\"name\":\"registry\"}],\"terminationGracePeriodSeconds\":5}},\"updateStrategy\":{\"type\":\"RollingUpdate\"}}}"
```
diafour, line 31: I think it is better to use object as is, without transforming it into a string. Also, we can support both variants.

KUBERNETES.md
```diff
@@ -0,0 +1,134 @@
...
+* `namespace` — object's Namespace. If empty, implies operation on a Cluster-level resource.
+* `name` — object's name.
+* `mergePatch` — describes transformations to perform on an object.
+* `subresource` — e.g., `status`.
+
+##### Example
+
+```json
+{
+  "operationType": "JQPatch",
```
diafour, line 105:
```suggestion
  "operationType": "MergePatch",
```

KUBERNETES.md
```diff
@@ -0,0 +1,134 @@
...
+* `namespace` — object's Namespace. If empty, implies operation on a Cluster-level resource.
+* `name` — object's name.
+* `jsonPatch` — describes transformations to perform on an object.
+* `subresource` — e.g., `status`.
+
+##### Example
+
+```json
+{
+  "operationType": "JQPatch",
```
diafour, line 128:
```suggestion
  "operationType": "JSONPatch",
```

KUBERNETES.md
```diff
@@ -0,0 +1,134 @@
...
+## Operations
+
+### Create
+
+* `operationType` — specifies an operation's type. Deletion types map directly to Kubernetes
+  DELETE's [`propagationPolicy`](https://kubernetes.io/docs/concepts/workloads/controllers/garbage-collection/).
+    * `CreateOrUpdate` — accept a Kubernetes object.
+      It retrieves an object, and if it already exists, computes a JSON Merge Patch and applies it (will not update .status field).
+      If it does not exist, we create the object.
+    * `Create` — will fail if an object already exists
```
diafour, line 21: Does this fail will lead to a hook error, so the hook will be restarted?

zuzzas: It will lead to the hook error.

KUBERNETES.md
zuzzas: >Also, we can support both variants.

This seems perfect. I'll change the doc to use a non-ugly object representation. :)

KUBERNETES.md
zuzzas: Seems reasonable.

KUBERNETES.md
```diff
@@ -0,0 +1,180 @@
...
+
+##### Example
+
+```json
+{
+  "operationType": "JSONPatch",
+  "kind": "Deployment",
+  "namespace": "default",
+  "name": "nginx",
+  "jsonPatch": "[{\"op\":\"replace\",\"path\":\"/spec/replicas\",\"value\":1}]"
```
nabokihms, line 178: Same thought about this field as in https://github.com/flant/shell-operator/pull/239#discussion_r548863035
Maybe we can use an object as the value of `jsonPatch` and `mergePatch` fields.

KUBERNETES.md
```diff
@@ -0,0 +1,180 @@
...
+#### JQPatch
+
+* `operation` — specifies an operation's type.
+* `apiVersion` — optional field that specifies object's apiVersion. If not present, we'll use preferred apiVersion
+  for the given kind.
+* `kind` — object's Kind.
+* `namespace` — object's Namespace. If empty, implies operation on a Cluster-level resource.
+* `name` — object's name.
+* `jqExpression` — describes transformations to perform on an object.
+* `subresource` — e.g., `status`.
```
nabokihms, line 122: Could you provide an example for this one? My question is whether I should write jq expression like `".status.lastError = (now | todate)"`, or similar to `".lastError = (now | todate)"`.

KUBERNETES.md
```diff
@@ -0,0 +1,180 @@
...
+
+### Delete
+
+* `operation` — specifies an operation's type. Deletion types map directly to Kubernetes
+  DELETE's [`propagationPolicy`](https://kubernetes.io/docs/concepts/workloads/controllers/garbage-collection/).
+  * `Delete` — foreground deletion. Hook will block the queue until the referenced object and all its descendants are deleted.
+  * `DeleteInBackground` — will delete the referenced object immediately. All its descendants will be removed by Kubernetes'
+    garbage collector.
+  * `DeleteNonCascading` — will delete the referenced object immediately, and orphan all its descendants.
+  apiVersion from `discovery` client.
```
nabokihms, line 89: Looks like an orphaned line :)

KUBERNETES.md
```diff
@@ -0,0 +1,180 @@
...
+* `apiVersion` — optional field that specifies object's apiVersion. If not present, we'll use preferred apiVersion
+  for the given kind.
+* `kind` — object's Kind.
+* `namespace` — object's namespace. If empty, implies operation on a cluster-level resource.
+* `name` — object's name.
+* `subresource` — e.g., `status`.
```
nabokihms, line 95: Do we really need to delete subresources? What will happen if we delete the status of a Deployment, for example?

KUBERNETES.md
```diff
@@ -0,0 +1,180 @@
...
+```
+
+### Patch
+
+Use `JQPatch` for almost everything. Consider using `MergePatch` or `JSONPatch` if you are attempting to modify 
+rapidly changing object, for example `status` field with many concurrent changes (and incrementing `resourceVersion`).
+
+#### JQPatch
+
+* `operation` — specifies an operation's type.
```
nabokihms, line 115: In a hook bindings configuration, there is an option called `allowFailure`. Maybe it can be useful for patches too.

KUBERNETES.md
zuzzas: Makes sense.

KUBERNETES.md
zuzzas: The first one. It'll go into the examples folder (or section).

KUBERNETES.md
zuzzas: Done.

KUBERNETES.md
zuzzas: Indeed.

KUBERNETES.md
zuzzas: The API allows us to send a delete request to a subresource. We aren't enforcing any other constraints.

pkg/hook/hook.go
```diff
@@ -110,6 +117,7 @@ func (h *Hook) Run(bindingType BindingType, context []BindingContext, logLabels
 		envs = append(envs, fmt.Sprintf("BINDING_CONTEXT_PATH=%s", contextPath))
 		envs = append(envs, fmt.Sprintf("METRICS_PATH=%s", metricsPath))
 		envs = append(envs, fmt.Sprintf("VALIDATING_RESPONSE_PATH=%s", validatingPath))
+		envs = append(envs, fmt.Sprintf("KUBERNETES_OBJECT_PATCH_SET_PATH=%s", objectPatchPath))
```
diafour, line 120: I'd rather name it "KUBERNETES_PATCH_PATH".

pkg/kube/client.go
```diff
@@ -180,6 +185,9 @@ func (c *kubernetesClient) Init() error {
 		return err
 	}
 
+	c.cachedDiscoveryClient = memory.NewMemCacheClient(c.Discovery())
+	go wait.Forever(func() { c.cachedDiscoveryClient.Invalidate() }, 2*time.Minute)
```
diafour, line 189: Great enhancement! I have some considerations:

1. Periodic invalidation is not always necessary.

- APIResourceList is called only on startup. So we can stop a periodic invalidation after all Monitors are started, e.g. if there are only conversion/validating hooks. 
- A constant period is not good, we've faced consequences already (see #226).

May be [JitterUntil](https://pkg.go.dev/k8s.io/apimachinery/pkg/util/wait#JitterUntil) or its buddy JitterUntilWithContext can give us more control over this cache mechanism.

2. Can we face a reincarnation of a problem with "non-working api controllers"? There is a NOTE in [docs](https://pkg.go.dev/k8s.io/client-go/discovery/cached/memory):
> NOTE: The client will NOT resort to live lookups on cache misses.

I can predict this scenario: "Invalidate" can't get GVRs → cache is empty → APIResourceList can't get GVR → shell-operator can't start informer for Pods.

I think we should elaborate more on this enhancement...

pkg/kube/object_patch/patch.go
```diff
@@ -0,0 +1,344 @@
...
+		if operationError != nil {
+			applyErrors = multierror.Append(applyErrors, operationError)
+		}
+	}
+
+	//goland:noinspection GoNilness
+	return applyErrors.ErrorOrNil()
+}
+
+func unmarshalFromJsonOrYaml(specs []byte) ([]OperationSpec, error) {
```
diafour, line 90: Do we need to support YAML responses from hooks?

test/integration/suite/run.go
```diff
@@ -26,7 +30,7 @@ func RunIntegrationSuite(t *testing.T, description string, clusterPrefix string)
 }
 
 var _ = SynchronizedBeforeSuite(func() []byte {
-	Ω(KindCreateCluster(ClusterName)).Should(Succeed())
```
diafour, line 29: So sad. Omega gives the invaluable feel of the esoteric knowledge ;]

test/utils/kind-k8s-cluster.go
```diff
@@ -112,8 +112,12 @@ func KindNodeImage() string {
...
 		"1.13": "kindest/node:v1.13.12",
 		"1.14": "kindest/node:v1.14.10",
-		"1.15": "kindest/node:v1.15.7",
-		"1.16": "kindest/node:v1.16.4",
+		"1.15": "kindest/node:v1.15.12",
+		"1.16": "kindest/node:v1.16.15",
+		"1.17": "kindest/node:v1.17.17",
+		"1.18": "kindest/node:v1.18.15",
+		"1.19": "kindest/node:v1.19.7",
+		"1.20": "kindest/node:v1.20.2",
```
diafour, line 120: I think we should bump "sigs.k8s.io/kind" to 0.10.0 and remove 1.11 - 1.13 here.

go.mod
```diff
@@ -20,6 +20,7 @@ require (
 	gopkg.in/alecthomas/kingpin.v2 v2.2.6
 	gopkg.in/robfig/cron.v2 v2.0.0-20150107220207-be2e0b0deed5
 	gopkg.in/satori/go.uuid.v1 v1.2.0
+	gopkg.in/yaml.v3 v3.0.0-20191120175047-4206685974f2
```
diafour, line 23: Consider using "sigs.k8s.io/yaml" instead.

pkg/kube/object_patch/patch_test.go
```diff
@@ -0,0 +1,67 @@
...
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			o := &ObjectPatcher{kubeClient: nil}
+			testSpecs := mustReadFile(tt.testFilePath)
+
+			_, err := o.ParseSpecs(testSpecs)
+			if (err != nil) != tt.wantErr {
+				t.Errorf("error = %v, wantErr %v", err, tt.wantErr)
```
nabokihms, line 63: As far as I know, we use gomega to make asserts in tests (or testify somewhere).

pkg/kube/object_patch/types.go
```diff
@@ -0,0 +1,30 @@
+package object_patch
+
+type OperationSpec struct {
+	Operation   OperationType `json:"operation" yaml:"operation"`
```
nabokihms, line 4: As we discussed internally, it is better to use `"sigs.k8s.io/yaml"`. And with `"sigs.k8s.io/yaml"` you only need to specify `json` tags.
https://play.golang.org/p/Vs4LQP3iL29

pkg/shell-operator/operator.go
```diff
@@ -538,8 +546,25 @@ func (op *ShellOperator) TaskHandleHookRun(t task.Task) queue.TaskResult {
 		err = op.HookMetricStorage.SendBatch(result.Metrics, map[string]string{
 			"hook": hookMeta.HookName,
 		})
+		if err != nil {
+			goto Metrics
```
nabokihms, line 550: Oh, we will definitely have to refactor this method in the future...

pkg/kube/object_patch/validation.go
```diff
@@ -0,0 +1,172 @@
+package object_patch
+
+import (
+	"k8s.io/apimachinery/pkg/util/validation/field"
+)
+
+func validateSpecs(specs []OperationSpec) error {
```
nabokihms, line 7: We use [openapi schema](https://github.com/flant/shell-operator/blob/master/pkg/hook/config/schemas.go#L11) to validate hooks configurations. Is it better to use the same approach here? WDYT?

pkg/kube/object_patch/patch.go
```diff
@@ -0,0 +1,344 @@
...
+	kind := object.GetKind()
+
+	gvk, err := o.kubeClient.GroupVersionResource(apiVersion, kind)
+	if err != nil {
+		return err
+	}
+
+	_, err = o.kubeClient.Dynamic().Resource(gvk).Namespace(namespace).Create(object, metav1.CreateOptions{}, generateSubresources(subresource)...)
+	if err != nil {
+		if errors.IsAlreadyExists(err) {
```
nabokihms, line 172: I think you don't have to check that `err != nil` before this check.

test/utils/kind-k8s-cluster.go
zuzzas: I've removed obsolete Kubernetes versions, but I can't bump the KIND version since it messes up dependencies.

pkg/shell-operator/operator.go
zuzzas: Aye, the control flow is horrible here. In my defense, it was horrible before I arrived. :)

pkg/kube/object_patch/validation.go
zuzzas: Oh, I was not aware of that. Yeah, openapi schema makes more sense. Declarative and everything in one place.

pkg/kube/object_patch/patch.go
zuzzas: Indeed!

pkg/kube/object_patch/patch_test.go
zuzzas: Assertions from these libraries are useless here. We validate whether the test table requires an error to not be nil on 62 and fail the test accordingly.

pkg/kube/object_patch/patch.go
```diff
@@ -0,0 +1,344 @@
...
+		case DeleteInBackground:
+			operationError = o.DeleteObjectInBackground(spec.ApiVersion, spec.Kind, spec.Namespace, spec.Name)
+		case DeleteNonCascading:
+			operationError = o.DeleteObjectNonCascading(spec.ApiVersion, spec.Kind, spec.Namespace, spec.Name)
+		case JQPatch:
+			operationError = o.JQPatchObject(spec.JQFilter, spec.ApiVersion, spec.Kind, spec.Namespace, spec.Name, spec.Subresource)
+		case MergePatch:
+			jsonMergePatch, err := json.Marshal(spec.MergePatch)
+			if err != nil {
+				applyErrors = multierror.Append(applyErrors, err)
```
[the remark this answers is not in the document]
zuzzas: Indeed!

pkg/kube/object_patch/patch.go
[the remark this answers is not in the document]
zuzzas: Done.

>Honestly, it doesn't matter

And for that, I am forever grateful.

zuzzas: Why not? Much easier to write by hand in simple hooks.

test/integration/suite/run.go
zuzzas: Just...

go.mod
zuzzas: @nabokihms 
How would I approach parsing multi-document YAML with a v2 library? The streaming `Decoder` interface is not supported there.

nabokihms: Something like this
https://github.com/kubernetes/apimachinery/blob/master/pkg/util/yaml/decoder.go

I think the only way, for now, is to use this code from the apimachinery library.

nabokihms: On the other hand, we can consider not using `sigs.k8s.io/yaml` :)

pkg/kube/client.go
zuzzas: Thanks for the list. I've referenced it (along with a link to this naïve implementation) in the [issue](https://github.com/flant/shell-operator/issues/254).

I've introduced this to possibly avoid useless API discovery when modifying Kubernetes objects. Clearly, the theoretical speed-up is not worth the hassle for now.

Let's scrap this feature for now since it affects the shell-operator in many ways I have not predicted.

pkg/hook/hook.go
zuzzas: Renamed!

test/utils/kind-k8s-cluster.go
zuzzas: https://github.com/flant/shell-operator/issues/255

go.mod
zuzzas: > we can consider not using `sigs.k8s.io/yaml`

You've referenced a fundamental difference between v2 (sigs) and v3 versions of the package. v3 package would fail in the following case:

```yaml
a: a
a: b
```

Are we dependent on this?

diafour: At the start, there was a ghodss/yaml, that was also used for addon-operator values. One moment I realize that the sigs.k8s.io/yaml is a successor of ghodss/yaml and it is a dependency of the client-go package.

Right now "sigs.k8s.io/yaml" is used to parse the hook configuration. This package can load "json-compatible" yaml (think map[string]interface{} instead of map[interface{}]interface{}) using json tags on structures. As JSON is a subset of YAML, the config is loaded directly as YAML https://github.com/flant/shell-operator/pull/239#discussion_r581820165.

At this time only JSON responses are supported from hooks. `NewDecoder` in json/encoding is used to parse "multi-document" or "streamed" JSON to support, for example, multiple metrics.

The feature to support multi-document YAML responses from hooks is a change of the protocol between shell-operator and a hook, that is why I'm asking "is it really needed?" [here](https://github.com/flant/shell-operator/pull/239#discussion_r579793228).

I think we should not change the protocol for one binging, but change it for all responses at once: metrics, conversion, validating.

So, fo me it looks like a separate PR where we can also migrate a config parser to yaml.v3.

pkg/kube/object_patch/types.go
zuzzas: We'll use yaml v3 with proper struct tags for each serialization format. More verbose, more correct.

go.mod
zuzzas: https://github.com/flant/shell-operator/issues/256

KUBERNETES.md
zuzzas: It's already taken care of [here](https://github.com/flant/shell-operator/blob/757a864d59bceed8ab6b52c1df22ba9f63a41869/pkg/shell-operator/operator.go#L709).

pkg/kube/object_patch/validation.go
zuzzas: OpenAPI schema is written. Spec parsing goes through schema validation now.

pkg/kube/object_patch/patch_test.go
nabokihms: This is not about saving the "if" statement or writing less code. Using the same matching library produces the same output style. For me, it looks cleaner to write something like in this particular case:
```go
if t.wantErr {
    g.Expect(err).To(HaveOccurred())
} else {
    g.Expect(err).ToNot(HaveOccurred())
}
```

pkg/kube/object_patch/validation.go
```diff
@@ -0,0 +1,199 @@
...
+		return nil, fmt.Errorf("expand schema: %v", err)
+	}
+
+	return s, nil
+}
+
+// See https://github.com/kubernetes/apiextensions-apiserver/blob/1bb376f70aa2c6f2dec9a8c7f05384adbfac7fbb/pkg/apiserver/validation/validation.go#L47
+func ValidateOperationSpec(obj interface{}, s *spec.Schema, rootName string) (multiErr error) {
+	if s == nil {
+		return fmt.Errorf("validate config: schema is not provided")
```
nabokihms, line 174: I don't think that the word "config" is correct for object patch validation (frankly, errors look like copypasted from another place). In my opinion, it would better to use specific terms like "object patch validation" or something like this.

nabokihms (approved):

diafour (changes requested):

pkg/kube/object_patch/patch.go
```diff
@@ -0,0 +1,355 @@
...
+			objCopy.SetResourceVersion(existingObj.GetResourceVersion())
+			_, err = o.kubeClient.Dynamic().Resource(gvk).Namespace(objCopy.GetNamespace()).Update(objCopy, metav1.UpdateOptions{}, generateSubresources(subresource)...)
+			return err
+		})
+	}
+
+	return err
+}
+
+func (o *ObjectPatcher) FilterObject(filterFunc func(*unstructured.Unstructured) (*unstructured.Unstructured, error),
```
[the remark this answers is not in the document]
zuzzas: For the Go hooks.

diafour (approved):

Issue #41, opened 2019-08-16
Full support for CRDs


distol:
Brief overview:
1. Add new kind of hooks for CR conversion requests.
2. Add cmd-line options for conversion webhook parameters (listen port, SSL, service, and so on).
3. Add a better way of installing CRDs (special directory; fill webhookClientConfig).

### Add new kind of hooks for CR conversion requests

1. Introduce new binding:
```
"onCustomResourceConversion": {
  "kind": "MyCustomResourceType",
  "conversions": [
    {from: "v1beta1", to: "v1beta2"}
  ]
}
```
2. TODO: Introduce format for passing ConversionRequest to the hooks
  * File with objects to convert (path to the file in an environment variable)
  * Several objects of the same kind can be passed at a time
  * DesiredAPIVersion field as an environment variable 
3. TODO: Introduce format for passing ConversionResponse from the hooks
  * Result using return code
  * Stdout as an error message
  * File to save to converted objects (path to the file in an environment variable) 

### Add cmd-line options for conversion webhook parameters
  * Port to listen
  * Certificate to use
  * CA to set to webhookClientConfig
  * Service name, port, etc. to set to webhookClientConfig

### Add a better way of installing CRDs
1. Add support for the directory with CRDs (exact path can be supplied using cmd-line args)
2. Parse all the CRDs in that directory and populate them with webhookClientConfig, if there are `onCustomResourceConversion` hooks, mentioning the same kind.
3. Install and update parsed and enriched CRDs to Kubernetes
4. Consider implementing preprocessing (using go-templates and sprig). Add built-in objects containing information about kubernetes version and available capabilities:
    - before 1.11 we don't have versions field, only version
    - before 1.13 we don't have webhook conversions (and until 1.15 they feature gate should be enabled)
    - before 1.15 we don't have defaulting

### Implement extra logic
1. Support automatic chaining of several conversion hooks, if needed. For example, if one hook can convert from v1alfa1 to v1alfa2, and another can do the conversion from v1alfa1 to v1beta1, and the api-server asks us to do the conversion from v1alfa1 to v1beta1 – to hooks has to be chained.
2. Consider implementing logic, that automatically finds all objects stored in an old apiVersion and triggers a change to rewrite them (and trigger conversion).
3. Add cmd-line command to delete all installed CRDs. By default, only CRDs having no objects should be deleted. Add `--force` flag, to also delete CRDs having stored objects.

distol:
```
kubernetesCustomResourceConversion:
[...]

PR #250, opened 2021-02-05
feat: support for conversion webhooks

diafour:
#### Overview

Use hooks as conversion webhooks.

See #41. 

This PR adds a new binding `kubernetesCustomResourceConversion`.

```release-note

```


pkg/shell-operator/operator.go
```diff
@@ -359,13 +363,139 @@ func (op *ShellOperator) InitWebhookManager() (err error) {
...
+		}
+		convPath := op.HookManager.FindConversionChain(event.CrdName, rule)
+		if len(convPath) == 0 {
+			continue
+		}
+		logEntry.Infof("Find conversion path for %s: %v", rule.String(), convPath)
+
+		for _, ruleID := range convPath {
+			var tasks []task.Task
+			op.HookManager.HandleConversionEvent(event, ruleID, func(hook *hook.Hook, info controller.BindingExecutionInfo) {
```
nabokihms, line 429: Can we move the code below to a separate method/function to reduce complexity?

pkg/webhook/conversion/chain.go
```diff
@@ -0,0 +1,251 @@
+package conversion
+
+import (
+	"strings"
+
+	"github.com/flant/shell-operator/pkg/utils/string_helper"
+)
+
+type ChainStorage map[string]*Chain
```
nabokihms, line 9: It is better to make ChainStorage a structure with an internal field for the map to make it more flexible.
For example, with the current approach, we can't add a simple lock to protect Chains.

pkg/webhook/conversion/chain.go
```diff
@@ -0,0 +1,251 @@
...
+		for id, path := range newPaths {
+			chain.PathsCache[id] = path
+		}
+	}
+
+	return nil
+}
+
+func (c Chain) SearchPathForRule(rule ConversionRule) []string {
+	IDs := []string{}
```
nabokihms, line 115: `ids`

pkg/webhook/conversion/config.go
```diff
@@ -0,0 +1,47 @@
...
+
+func (r ConversionRule) ShortFromVersion() string {
+	return string_helper.TrimGroup(r.FromVersion)
+}
+
+func (r ConversionRule) ShortToVersion() string {
+	return string_helper.TrimGroup(r.ToVersion)
+}
+
+func RuleFromString(in string) ConversionRule {
```
nabokihms, line 41: Looks strange. If I got it right, this is our internal format. Why we have to convert rules to strings and vice-versa?

pkg/webhook/server/server.go
```diff
@@ -62,7 +62,7 @@ func (s *WebhookServer) Start() error {
 		tlsConf.ClientCAs = roots
 	}
 
-	listenAddr := app.ValidatingWebhookSettings.ListenAddr + ":" + app.ValidatingWebhookSettings.ListenPort
+	listenAddr := s.Settings.ListenAddr + ":" + s.Settings.ListenPort
```
nabokihms, line 65: net package contains precious function `JoinHostPort` for this purpose :)

pkg/webhook/server/server.go
```diff
@@ -85,7 +85,6 @@ func (s *WebhookServer) Start() error {
 		err := srv.ServeTLS(listener, "", "")
 		if err != nil {
 			log.Errorf("Error starting Webhook https server: %v", err)
-			//os.Exit(1)
```
nabokihms, line 88: What will happen if the server crashes? I think it is better to restart the shell-operator in this case. There are two ways to do it:
1. Shutdown the shell-operator if one of the webhooks crashes.
2. Provide some sort of health check endpoint to ensure that everything works.

pkg/webhook/validating/handler.go
```diff
@@ -123,20 +123,21 @@ func (h *WebhookHandler) HandleReviewRequest(path string, body []byte) (*v1.Admi
...
-		if configurationId == "" {
-			configurationId = p
+		if configurationID == "" {
+			configurationID = p
 			continue
 		}
 		webhookParts = append(webhookParts, p)
 	}
-	webhookId = strings.Join(webhookParts, "/")
+	webhookID = strings.Join(webhookParts, "/")
```
nabokihms, line 140: Using `path` is more accurate.

examples/204-validating-webhook/README.md
```diff
@@ -1,6 +1,6 @@
 # Example with validating hooks
 
-This is a simple example of `kubernetesValidating` binding. Read more information in [VALIDATING.md](../../VALIDATING.md).
+This is a simple example of `kubernetesValidating` binding. Read more information in [BINDING_VALIDATING.md](../../BINDING_VALIDATING.md).
```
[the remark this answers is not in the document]
diafour: This branch is not rebased.

pkg/webhook/validating/handler.go
diafour: This webhookID is used as a key later in hook_manager, so the separator string should be constant.

pkg/webhook/server/server.go
diafour: Ouch, I need to return this back.

pkg/webhook/conversion/config.go
diafour: Strings are used to quickly implement a conversion path searching algorithm.

pkg/webhook/conversion/chain.go
diafour: Right now lock is not required for ChainStorage — it is filled once at startup. But it should be a good idea to add a lock for maps in Chain structure.

pkg/hook/testdata/hook_manager_conversion_chains/hook.sh
```diff
@@ -0,0 +1,30 @@
...
+
+if [[ $1 == "--config" ]] ; then
+    cat <<EOF
+configVersion: v1
+onStartup: 10
+kubernetesCustomResourceConversion:
+- name: up_conversions
+  crdName: crontabs.stable.example.com
+  conversions:
+  - fromVersion: azaza
```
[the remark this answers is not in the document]
diafour: LOL

pkg/webhook/validating/handler.go
nabokihms: That is exactly what the path package provides.

examples/204-validating-webhook/README.md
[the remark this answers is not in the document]
nabokihms: Can you rebase it then?

pkg/webhook/conversion/chain.go
nabokihms: A lock is just an example. We can not add new properties to the map, but we add anything we want to the struct.

pkg/webhook/validating/handler.go
diafour: > // Package path implements utility routines for manipulating slash-separated
// paths.

https://github.com/golang/go/blob/master/src/path/path.go#L5-L6

I thought that `path` uses a system-dependent separator. Anyway, path.Join does a huge redundant [Clean](https://github.com/golang/go/blob/master/src/path/path.go#L70). I would rather use SplitN and drop append+Join.

nabokihms: Ok, that sounds reasonable.

pkg/shell-operator/operator.go
nabokihms: We decided to leave it as is.

examples/204-validating-webhook/README.md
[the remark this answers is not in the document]
diafour: Done

examples/210-conversion-webhook/templates/certs-secret.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: v1
+kind: Secret
+metadata:
+  name: example-210-conversion-certs
+type: Opaque
```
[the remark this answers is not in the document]
diafour: Done! Also fixed in 204-validating-webhook example.

pkg/webhook/conversion/chain.go
diafour: Done. No more string ids, Rule struct is used as key.

pkg/webhook/conversion/config.go
diafour: Refactored.

pkg/webhook/server/server.go
diafour: Done. JoinHostPort is used now.

pkg/webhook/server/server.go
diafour: Reverted!

pkg/hook/config/config.go
```diff
@@ -0,0 +1,155 @@
...
+import (
+	"fmt"
+
+	"sigs.k8s.io/yaml"
+
+	. "github.com/flant/shell-operator/pkg/hook/types"
+)
+
+// HookConfig is a structure with versioned hook configuration
+type HookConfig struct {
```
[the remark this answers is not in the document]
diafour: It also can be "Bindings" or "HookBindings". I'll leave it as is for now until a better naming comes up.

diafour:
Ready: reviewed and tested many times internally.

PR #252, opened 2021-02-09
frameworks and shell-operator already in base image

arve0:
#### Overview
Do not add files from context that are already present in base image `flant/shell-operator:latest`.

Currently the build fails with:
```
Step 3/5 : ADD frameworks/shell /frameworks/shell
ADD failed: file not found in build context or excluded by .dockerignore: stat frameworks/shell: file does not exist
```

No.


diafour:
Hello! That's a development commands, what a shame 🤦 . Also, there are some fixes for 204-validating-webhook in  https://github.com/flant/shell-operator/pull/250.

PR #253, opened 2021-02-15
tag latest-alpine3.11 not present in docker hub

arve0:
#### Overview
Valid docker tag in exmaples.

Current tag is not present in docker hub.

No.


diafour:
Oh! Thanks!

PR #257, opened 2021-03-01
feat: execution rate settings for hook

diafour:
#### Overview

Introduce settings to set execution rate for a hook.

Shell-operator can combine tasks in the queue and run the hook once with an array of binding contexts. Execution rate settings help to handle multiple events that are occurred during a short period.

Yes, the new section "settings" is added to the configuration.

```release-note

```

### DIFF
--- a/pkg/hook/hook.go
+++ b/pkg/hook/hook.go
@@ -1,6 +1,7 @@
 package hook
 
 import (
+	"context"
 	"fmt"
 	"io/ioutil"
 	"os"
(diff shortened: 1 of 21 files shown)


pkg/hook/hook.go
```diff
@@ -260,3 +271,14 @@ func (h *Hook) prepareConversionResponseFile() (string, error) {
 
 	return conversionPath, nil
 }
+
+func CreateRateLimiter(cfg *config.HookConfig) *rate.Limiter {
```
nabokihms, line 275: What do you think about returning nil from this method and then calling `Wait` only if RateLimitter is not `nil`? It will help us to avoid unnecessary interactions.

diafour: I thought about it. The default RateLimiter has "Inf" limit, so `Wait` returns immediately. We can hide limiter inside RateLimitWait (https://github.com/flant/addon-operator/pull/176/files#r586508346) and do `nil` checking there.

zuzzas (approved):

nabokihms (approved):
Lgtm

PR #266, opened 2021-03-15
Set global request timeout for Kubernetes client

zuzzas:
#### Overview

Set global request timeout for Kubernetes client

Some operations (especially in the "object_patch" package) may get stuck for an inordinate amount of time.

```release-note

```


diafour:
Nice workaround before update client-go to 1.20!

PR #267, opened 2021-03-16
Send empty interface{} to the filterFunc

zuzzas:

We want typed filter results for Go hooks.

Related PR in addon-operator: https://github.com/flant/addon-operator/pull/180

There are no changes in shell hooks configuration.

```release-note

```

### DIFF
--- a/pkg/kube_events_manager/filter.go
+++ b/pkg/kube_events_manager/filter.go
@@ -18,7 +18,7 @@ import (
 // ApplyFilter filters object json representation with jq expression, calculate checksum
 // over result and return ObjectAndFilterResult. If jqFilter is empty, no filter
 // is required and checksum is calculated over full json representation of the object.
-func ApplyFilter(jqFilter string, filterFn func(obj *unstructured.Unstructured) (result string, err error), obj *unstructured.Unstructured) (*ObjectAndFilterResult, error) {
+func ApplyFilter(jqFilter string, filterFn func(obj *unstructured.Unstructured) (result interface{}, err error), obj *unstructured.Unstructured) (*ObjectAndFilterResult, error) {
 	defer trace.StartRegion(context.Background(), "ApplyJqFilter").End()
 
 	res := &ObjectAndFilterResult{
@@ -31,18 +31,24 @@ func ApplyFilter(jqFilter string, filterFn func(obj *unstructured.Unstructured) 
 	if err != nil {
 		return nil, err
 	}
-	res.ObjectBytes = int64(len(data))
+	res.ObjectSize = len(data)
 
 	// If filterFn is passed, run it and return result.
 	if filterFn != nil {
-		var err error
-		var filtered string
-		filtered, err = filterFn(obj)
+		filteredObj, err := filterFn(obj)
 		if err != nil {
 			return nil, fmt.Errorf("filterFn: %v", err)
 		}
-		res.FilterResult = filtered
-		res.Metadata.Checksum = utils_checksum.CalculateChecksum(filtered)
+
+		filteredBytes, err := json.Marshal(filteredObj)
+		if err != nil {
+			return nil, err
+		}
+
+		res.FilterResult = filteredObj
+		res.FilterResultSize = len(filteredBytes)
+		res.Metadata.Checksum = utils_checksum.CalculateChecksum(string(filteredBytes))
+
 		return res, nil
 	}
 
@@ -55,8 +61,11 @@ func ApplyFilter(jqFilter string, filterFn func(obj *unstructured.Unstructured) 
 		if err != nil {
 			return nil, fmt.Errorf("jqFilter: %v", err)
 		}
+
 		res.FilterResult = filtered
+		res.FilterResultSize = len(filtered)
 		res.Metadata.Checksum = utils_checksum.CalculateChecksum(filtered)
 	}
+
 	return res, nil
 }
--- a/pkg/kube_events_manager/types/types.go
+++ b/pkg/kube_events_manager/types/types.go
@@ -6,7 +6,6 @@ import (
 	"strings"
 
 	log "github.com/sirupsen/logrus"
-
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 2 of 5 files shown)


pkg/kube_events_manager/types/types.go
```diff
@@ -80,17 +74,29 @@ func (o ObjectAndFilterResult) MarshalJSON() ([]byte, error) {
...
 
 type ObjectAndFilterResults map[string]*ObjectAndFilterResult
 
-func (a ObjectAndFilterResults) Bytes() (size int64) {
+func (a ObjectAndFilterResults) Bytes() (size int) {
 	for _, o := range a {
-		size += int64(len(o.FilterResult))
-		size += o.ObjectBytes
+		if newSize, overflowed := overflow.Add(size, o.FilterResultSize); overflowed {
+			size = mathutil.MaxInt
```
diafour, line 86: I don't get it, why use int instead of int64?

pkg/kube_events_manager/types/types.go
```diff
@@ -57,19 +58,12 @@ func (o ObjectAndFilterResult) Map() map[string]interface{} {
...
-	inJson := o.FilterResult
-	if inJson == "" {
-		m["filterResult"] = nil
-		return m
-	}
-	var res interface{}
-	err := json.Unmarshal([]byte(inJson), &res)
-	if err != nil {
-		log.Errorf("Possible bug!!! Cannot unmarshal jq filter '%s' result: %s", o.Metadata.JqFilter, err)
+	if o.FilterResult == nil {
```
diafour, line 61: As I understand, this method is not used for go hooks, so FilterResult here can't be nil for jqFilter:

`res.FilterResult = []byte(filtered)`

https://github.com/flant/shell-operator/pull/267/files#diff-37508d544287271c578df624603d224bf1332acbb608e7c5dbfeea23e9ac72a0R70

diafour (changes requested):

pkg/kube_events_manager/types/types.go
zuzzas: Reverted to int64.

pkg/kube_events_manager/types/types.go
zuzzas: Fixed this. It is now Umarshalled from JSON, exactly as it was before my changes. If there was a `null` inside supposed JSON string, then it'll be nil.

test/hook/context/generator.go
```diff
@@ -54,7 +54,7 @@ func convertBindingContexts(bindingContexts []BindingContext) (GeneratedBindingC
 		return res, fmt.Errorf("marshaling binding context error: %v", err)
 	}
 
-	res.BindingContexts = bindingContexts
+	res.BindingContexts = compactedBindingContexts
```
zuzzas, line 57: This was a bug that skipped compacting contexts when using Go hooks in tests.

diafour:
lgtm!

pkg/kube_events_manager/types/types.go
```diff
@@ -57,13 +60,19 @@ func (o ObjectAndFilterResult) Map() map[string]interface{} {
 		return m
 	}
 	// Add filterResult field only if it was requested
-	inJson := o.FilterResult
-	if inJson == "" {
+	filterResString, ok := o.FilterResult.(string)
+	if !ok {
```
nabokihms, line 64:
```suggestion
	if !ok || filterResString == "" {
```

pkg/kube_events_manager/types/types.go
```diff
@@ -80,17 +89,29 @@ func (o ObjectAndFilterResult) MarshalJSON() ([]byte, error) {
...
 }
 
 type ObjectAndFilterResults map[string]*ObjectAndFilterResult
 
 func (a ObjectAndFilterResults) Bytes() (size int64) {
 	for _, o := range a {
-		size += int64(len(o.FilterResult))
-		size += o.ObjectBytes
+		if newSize, overflowed := overflow.Add64(size, int64(o.FilterResultSize)); overflowed {
+			size = math.MaxInt64
```
nabokihms, line 101: In case of overflow, we lose precision. It looks like we need to add some warning here or even to return an error. WDYT?

pkg/kube_events_manager/filter.go
```diff
@@ -31,18 +31,24 @@ func ApplyFilter(jqFilter string, filterFn func(obj *unstructured.Unstructured)
...
 		}
-		res.FilterResult = filtered
-		res.Metadata.Checksum = utils_checksum.CalculateChecksum(filtered)
+
+		filteredBytes, err := json.Marshal(filteredObj)
+		if err != nil {
+			return nil, err
+		}
+
+		res.FilterResult = filteredObj
```
nabokihms, line 48: Just curious. Do we have a test for filter func and its memory measurement?

pkg/kube_events_manager/types/types.go
diafour: Although exabyte-scaled binding contexts are possible, I think the shell-operator should fail before counting their size =)

pkg/kube_events_manager/filter.go
zuzzas: There isn't a test for jqFilter or filterFunc.

pkg/kube_events_manager/types/types.go
zuzzas: I've obliterated the overflow test. Why? Well, we are losing even more precision when we export metrics by converting them to float64. I'd rework this calculation in v2 and skip it completely for now.

image

nabokihms (approved):
No concerns left.

pkg/kube_events_manager/types/types.go
```diff
@@ -57,13 +58,19 @@ func (o ObjectAndFilterResult) Map() map[string]interface{} {
...
+	filterResString, ok := o.FilterResult.(string)
+	if !ok || filterResString == "" {
 		m["filterResult"] = nil
 		return m
 	}
+
+	if filterResString == "" {
+		m["filterResult"] = nil
+		return m
+	}
```
diafour, line 70: This block is now redundant after the last commit =)

```suggestion
```

PR #268, opened 2021-03-17
Cached discovery and object patch debug logging

zuzzas:
#### Overview

1. Implemented disk-backed cached discovery for all Kubernetes clients in the shell-operator.
2. Added comprehensive debug logging to the `object_patch` package.

Object patch worked extremely slow since we had to discover resources on each patch operation.

1. I've peered into disk cache implementation, and it works much better than the memory cache. It is used by almost every controller in the `k8s.io/kubernetes` repo.
2. It's not possible to specify jitter for the cached discovery client, but I believe that's unnecessary since it generates a couple of HTTP calls and deletes/writes some files to the page cache.

```release-note

```


pkg/kube/client.go
```diff
@@ -395,6 +411,13 @@ func (c *kubernetesClient) GroupVersionResource(apiVersion string, kind string)
 
 }
 
+func (c *kubernetesClient) discovery() discovery.DiscoveryInterface {
```
diafour, line 414: For the fake client.

diafour (approved):

PR #273, opened 2021-04-13
feat(tests): add "AllSnapshots" binding context generator

diafour:
#### Overview

- Add "AllSnapshots" binding context generator.

- Replace custom binding context combiner for tests with CombineBindingContextForHook from ShellOperator.

```release-note

```

### DIFF
--- a/test/hook/context/context_combiner.go
+++ b/test/hook/context/context_combiner.go
@@ -1,0 +1,97 @@
+package context
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/flant/shell-operator/pkg/hook/binding_context"
+	"github.com/flant/shell-operator/pkg/hook/controller"
+	"github.com/flant/shell-operator/pkg/hook/task_metadata"
+	"github.com/flant/shell-operator/pkg/hook/types"
+	shell_operator "github.com/flant/shell-operator/pkg/shell-operator"
+	"github.com/flant/shell-operator/pkg/task"
+	"github.com/flant/shell-operator/pkg/task/queue"
+)
+
+// TestTaskType is a type for tasks created by ContextCombiner.
+const TestTaskType task.TaskType = "TestTask"
+
+// TestQueueName is a name of the queue created by ContextCombiner.
+const TestQueueName string = "test-queue"
+
+// ContextCombiner is used to bring logic behind the CombineBindingContextForHook method
+// into tests. This method requires a ShellOperator instance with at least one queue
+// with tasks.
+type ContextCombiner struct {
+	op *shell_operator.ShellOperator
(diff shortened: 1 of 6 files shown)


test/hook/context/context_combiner.go
```diff
@@ -0,0 +1,90 @@
...
+type ContextCombiner struct {
+	op *shell_operator.ShellOperator
+	q  *queue.TaskQueue
+}
+
+func NewContextCombiner() *ContextCombiner {
+	op := &shell_operator.ShellOperator{}
+	op.TaskQueues = queue.NewTaskQueueSet()
+	op.TaskQueues.WithContext(context.Background())
+	op.TaskQueues.NewNamedQueue("main", nil)
```
nabokihms, line 27: I did not get why there is the "main" queue now. Previously we did not specify any queue. And it is better to make `"main"` constant like `"TestTask"`.

test/hook/context/context_combiner.go
```diff
@@ -0,0 +1,90 @@
...
+	t := task.NewTask(TestTaskType)
+	t.WithMetadata(task_metadata.HookMetadata{
+		Binding:        info.Binding,
+		Group:          info.Group,
+		BindingType:    bindingType,
+		BindingContext: info.BindingContext,
+		AllowFailure:   info.AllowFailure,
+	})
+	t.WithQueueName("main")
+	c.q.AddLast(t)
```
nabokihms, line 45: Oh, we use queue only to add tasks? Ok then.

test/hook/context/context_combiner.go
```diff
@@ -0,0 +1,90 @@
...
+// CombinedContext returns a combined context or a binding context
+// from the first task - as a Shell-operator will do.
+func (c *ContextCombiner) Combined() []binding_context.BindingContext {
+	firstTask := c.q.GetFirst()
+	if firstTask == nil {
+		return []binding_context.BindingContext{}
+	}
+	taskMeta := firstTask.GetMetadata()
+
+	bc := taskMeta.(task_metadata.BindingContextAccessor).GetBindingContext()
```
nabokihms, line 57: Why do we need to cast a type here?

nabokihms (approved):
Approved in advance. The only concern is to move the default queue name to constant.

test/hook/context/context_combiner.go
diafour: Metadata is interface{} to have different HookMetadata in shell-operator and in addon-operator.

test/hook/context/context_combiner.go
diafour: I've added more comments and created a constant for queue's name.

PR #280, opened 2021-05-06
Add histogram support for custom (non-grouped) metrics

dulvac:
#### Overview

These changes add support for using the observe action on custom metrics from hooks

For custom metrics, we have the "add", "set" (and "expire") actions available in the hook, which correspond to counters and gauges respectively. In order to observe a duration from inside the hook into a custom metric, this PR adds support for the "observe" action. This is useful for wanting to observe durations for different parts of the hook's code.

Additionally, there are a few unrelated minor changes (increasing durations of buckets to cover usecases of longer-running hooks for "core" metrics), .gitignore and a couple of typos

```release-note
Add support for observing a histogram duration in custom, non-grouped metrics

```

### DIFF
--- a/pkg/metric_storage/metric_storage.go
+++ b/pkg/metric_storage/metric_storage.go
@@ -17,9 +17,6 @@ import (
 )
 
 const (
-	HistogramDefaultStart float64 = 0.0
-	HistogramDefaultWidth float64 = 2
-	HistogramDefaultCount int     = 40
 	PrefixTemplate                = "{PREFIX}"
 )
 
@@ -256,9 +253,6 @@ func (m *MetricStorage) Histogram(metric string, labels map[string]string, bucke
 }
 
 func (m *MetricStorage) RegisterHistogram(metric string, labels map[string]string, buckets []float64) *prometheus.HistogramVec {
-	if m != nil && len(buckets) != 0 {
-		m.HistogramBuckets[metric] = buckets
-	}
 	metricName := m.ResolveMetricName(metric)
 
 	defer func() {
@@ -278,10 +272,16 @@ func (m *MetricStorage) RegisterHistogram(metric string, labels map[string]strin
 
 	log.WithField("operator.component", "metricsStorage").
 		Infof("Create metric histogram %s", metricName)
-	buckets, has := m.HistogramBuckets[metric]
-	if !has {
-		buckets = prometheus.LinearBuckets(HistogramDefaultStart, HistogramDefaultWidth, HistogramDefaultCount)
+	b, has := m.HistogramBuckets[metric]
+	// This shouldn't happen except when entering this concurrently
+	// If there are buckets for this histogram about to be registered, keep them
+	// Otherwise, use the new buckets.
+	// No need to check for nil or empty slice, as the p8s lib will use DefBuckets
+	//(https://pkg.go.dev/github.com/prometheus/client_golang/prometheus#HistogramOpts)
+	if has {
+		buckets = b
 	}
+
 	vec = prometheus.NewHistogramVec(prometheus.HistogramOpts{
 		Name:    metricName,
 		Help:    metricName,
@@ -366,8 +366,16 @@ func (m *MetricStorage) ApplyOperation(op operation.MetricOperation, commonLabel
 		m.CounterAdd(op.Name, *op.Value, labels)
 		return
 	}
+	if op.Add != nil {
+		m.CounterAdd(op.Name, *op.Add, labels)
+		return
+	}
 	if op.Action == "set" && op.Value != nil {
 		m.GaugeSet(op.Name, *op.Value, labels)
+		return
+	}
+	if op.Set != nil {
+		m.GaugeSet(op.Name, *op.Set, labels)
 		return
 	}
 	if op.Action == "observe" && op.Value != nil && op.Buckets != nil {
@@ -390,8 +398,14 @@ func (m *MetricStorage) ApplyGroupOperations(group string, ops []operation.Metri
 		if op.Action == "add" && op.Value != nil {
 			m.GroupedVault.CounterAdd(group, op.Name, *op.Value, labels)
 		}
+		if op.Add != nil {
+			m.GroupedVault.CounterAdd(group, op.Name, *op.Add, labels)
+		}
 		if op.Action == "set" && op.Value != nil {
 			m.GroupedVault.GaugeSet(group, op.Name, *op.Value, labels)
+		}
+		if op.Set != nil {
+			m.GroupedVault.GaugeSet(group, op.Name, *op.Set, labels)
 		}
 	}
 }
--- a/pkg/metric_storage/metric_storage.go
+++ b/pkg/metric_storage/metric_storage.go
@@ -232,11 +232,7 @@ func (m *MetricStorage) RegisterCounter(metric string, labels map[string]string)
 }
 
 // Histograms
-func (m *MetricStorage) HistogramObserve(metric string, value float64, labels map[string]string) {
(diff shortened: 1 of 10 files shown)


pkg/metric_storage/metric_storage.go
```diff
@@ -369,14 +378,17 @@ func (m *MetricStorage) SendBatch(ops []operation.MetricOperation, labels map[st
...
+		m.CounterAdd(op.Name, *op.Value, labels)
 		return
 	}
-	if op.Set != nil {
-		m.GaugeSet(op.Name, *op.Set, labels)
+	if op.Action == "set" && op.Value != nil {
+		m.GaugeSet(op.Name, *op.Value, labels)
 		return
 	}
+	if op.Action == "observe" && op.Value != nil {
```
nabokihms, line 389: Could you add an example for the "observe" action to [the doc](https://github.com/flant/shell-operator/blob/master/METRICS.md#custom-metrics)?

nabokihms (commented):
Hello! Thank you for the contribution. Overall, it looks great! I would like @diafour to do a review.

pkg/metric_storage/metric_storage.go
dulvac: @nabokihms, done! Is that example ok?

dulvac:
> Hello! Thank you for the contribution. Overall, it looks great!
@nabokihms, you're welcome. Thanks for your quick responses.

pkg/metric_storage/metric_storage.go
nabokihms: Good enough, thanks.

diafour:
Great job @dulvac, thank you!

**Custom histograms**

Here are some thoughts about custom histograms:

- `LinearBuckets`, `HistogramDefaultCount` with others are the first implementation for Histograms. Since then, we use explicit buckets for Histogram registration. This "default" bucket initialization with linear spread appears to be not helpful. You can drop it.
- As we use explicit buckets, `RegisterHistogram`, `RegisterHistogramWithBuckets` and `HistogramDefineBuckets` should be replaced with one method `RegisterHistogram(metric string, labels map[string]string, buckets []float64)`.
- `"buckets"` should be a required field if hook returns operation with `"observe"` field.
- It'll be good to support grouped histograms or add an "unsupported" warning in the documentation.

**Increased buckets for shell-operator's histograms**

Additional buckets in shell-operator metrics are a tricky part: more information on the one side and more bytes on the other side. Each histogram generates buckets count + 2 timeseries, and this count is then multiplied by hooks count. It's not a big deal for simple cases, but multiple hook installations can be affected.

As a compromise, I propose these changes:
- "tasks_queue_action_duration_seconds" - 1,2,5,10,20 microseconds can de deleted and 10,20,50,100,200 milliseconds can be added. Real metrics indicate there is lack of information about millisecond ranges.
- "kube_jq_filter_duration_seconds" - this one measures time of jq filter applying for *one* Kubernetes object. Normally, it should not exceed 1 sec. We have rare occasions of 2-5 second durations for huge Node objects. I think 20,50,100,200,500 seconds are not applicable here.
- "kube_event_duration_seconds" - it is "kube_jq_filter_duration_seconds" plus cache operations.
- "hook_run_seconds", "hook_run_user_cpu_seconds", "hook_run_sys_cpu_seconds" - here we can drop 1,2,5,10 millisecond buckets. Real metrics have no information in these buckets *even for Go hooks*.

@nabokihms @zuzzas, what do you think about new buckets for shell-operator's histograms?

dulvac:
> Additional buckets in shell-operator metrics are a tricky part: more information on the one side and more bytes on the other side. Each histogram generates buckets count + 2 timeseries, and this count is then multiplied by hooks count. It's not a big deal for simple cases, but multiple hook installations can be affected.

@diafour , I see your point and it makes total sense. These will never be _right_ for any use case, where the duration depends on the hooks, which can be fast or longer-running. Maybe eventually we add a configuration file for some of these buckets, but this seems overkill as well. 

> "tasks_queue_action_duration_seconds" - 1,2,5,10,20 microseconds can de deleted and 10,20,50,100,200 milliseconds can be added. Real metrics indicate there is lack of information about millisecond ranges.

sure, makes sense.

> "kube_jq_filter_duration_seconds" - this one measures time of jq filter applying for one Kubernetes object. Normally, it should not exceed 1 sec. We have rare occasions of 2-5 second durations for huge Node objects. I think 20,50,100,200,500 seconds are not applicable here.

I think that was an accident on my part. will correct.

> "hook_run_seconds", "hook_run_user_cpu_seconds", "hook_run_sys_cpu_seconds" - here we can drop 1,2,5,10 millisecond buckets. Real metrics have no information in these buckets even for Go hooks.

great! Totally agree. (FYI, we are using go hooks, but they're dealing with the apiserver and external resources so yeah... definitely not relevant except maybe for no-ops, but those take longer as well.)

dulvac:
@diafour , @nabokihms , please have another look. I implemented the feedback around:
* shell-operator histogram buckets
* Registering histograms with buckets. FYI, the prometheus library will give you a DefBuckets slice if nil is passed, but that's validated now to not be the case for custom histograms. Also, I went down the rabbit hole a little bit, since `HistogramObserve` calls `Histogram`, which calls `RegisterHistogram` if the histogram is not already registered, which now needs buckets; so `HistogramObserve` now also needs buckets.

Let me know if it's ok.

pkg/metric_storage/metric_storage.go
```diff
@@ -369,14 +362,17 @@ func (m *MetricStorage) SendBatch(ops []operation.MetricOperation, labels map[st
...
-	if op.Add != nil {
-		m.CounterAdd(op.Name, *op.Add, labels)
+	if op.Action == "add" && op.Value != nil {
+		m.CounterAdd(op.Name, *op.Value, labels)
 		return
 	}
-	if op.Set != nil {
-		m.GaugeSet(op.Name, *op.Set, labels)
+	if op.Action == "set" && op.Value != nil {
+		m.GaugeSet(op.Name, *op.Value, labels)
```
diafour, line 370: It is better to check both conditions.

```suggestion
	if op.Set != nil {
		m.GaugeSet(op.Name, *op.Set, labels)
		return
	}
	if op.Action == "set" && op.Value != nil {
		m.GaugeSet(op.Name, *op.Value, labels)
```

pkg/metric_storage/operation/operation.go
```diff
@@ -111,11 +115,11 @@ func ValidateMetricOperation(op MetricOperation) error {
 	var opErrs *multierror.Error
 
 	if op.Action == "" {
-		opErrs = multierror.Append(opErrs, fmt.Errorf("one of: 'action', 'set' or 'add' is required: %s", op))
+		opErrs = multierror.Append(opErrs, fmt.Errorf("one of: 'action', 'set', 'add' or 'observe' is required: %s", op))
```
diafour, line 118: The original message is correct: the 'observe' shortcut is not implemented by this PR.

diafour (changes requested):
@dulvac some last changes are required.

A couple of words about 'set' and 'add' — these shortcuts are from the initial version of custom metrics. They become obsolete with "action" and "value" but still supported. I have a feeling, that we still have Go hooks that use these fields.

dulvac:
@diafour , thanks for the reviews! 
Please have another look.

nabokihms:
@dulvac Ah, sorry for the delay. @diafour is on vacation right now, but he will be available in the next week.

diafour (approved):

diafour:
Tests are good, so I think it's time to merge! Thank you @dulvac!

dulvac:
Thanks guys, it was a pleasure! Till the next one.

PR #282, opened 2021-05-06
Fake cluster rework

yalosev:

```release-note

```


test/hook/context/generator.go
```diff
@@ -33,17 +30,19 @@ type BindingContextController struct {
...
-	FakeCluster = fake.NewFakeCluster()
-
 	return &BindingContextController{
 		HookMap:    make(map[string]string),
 		HookConfig: config,
 
 		Controller:    NewStateController(),
 		UpdateTimeout: 1500 * time.Millisecond,
+
+		fakeCluster: fake.NewFakeCluster(fake.ClusterVersionV119),
```
diafour, line 45: It seems that BindingContextController always uses 1.19 and there is no way to change this version in tests.

pkg/kube/fake/scripts/resources_generator
```diff
@@ -0,0 +1,116 @@
+#!/bin/bash
```
zuzzas, line 1: How do you start a test cluster for each Kubernetes version? Can't see it anywhere in this PR.

zuzzas (changes requested):

pkg/kube/fake/scripts/resources_generator
yalosev: You can find description [here](https://github.com/flant/shell-operator/blob/fake-cluster-rework/pkg/kube/fake/resources.go#L6)

Shortly: It's a one time operation. We can use any cluster to take resources manually

test/hook/context/generator.go
```diff
@@ -33,17 +30,23 @@ type BindingContextController struct {
...
-}
 
-func NewBindingContextController(config string) (*BindingContextController, error) {
-	FakeCluster = fake.NewFakeCluster()
+	fakeCluster *fake.FakeCluster
+}
 
+func NewBindingContextController(config string, version ...fake.ClusterVersion) (*BindingContextController, error) {
+	k8sVersion := fake.ClusterVersionV119
+	if len(version) > 0 {
```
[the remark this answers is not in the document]
yalosev: dont think so, its a simple pattern for Go. Just drop 1+ values

test/hook/context/generator.go
yalosev: Fixed

pkg/kube/fake/scripts/resources_generator
zuzzas: Please, create an issue. We should have a way to get this from a clean `kind` cluster. That should be deployed via a script.

PR #288, opened 2021-05-14
Generate full resources not only preferred

yalosev:
#### Overview

Generated all resources for k8s fake cluster from 1.16 to 1.21 

Previously we had only preferred versions generated. It affects some clusters like 1.16
This PR has all resources generated

NONE


nabokihms (approved):
Looks good, thanks!

PR #289, opened 2021-05-18
Binding context controller fake cluster

yalosev:
#### Overview
Fix fake cluster for hook testing

 - Fix k8s resources
 - Migrate client-go lib to 0.19.11
 - Avoid fake cluster problems with Preferred resource versions
 - Fix pluralizing and snapshot generation

All this changes lead to more comfortable hook tests based on fake cluster resources

None


pkg/kube/fake/cluster.go
```diff
@@ -1,17 +1,19 @@
...
 
-	"github.com/flant/shell-operator/pkg/utils/manifest"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/version"
 	fakediscovery "k8s.io/client-go/discovery/fake"
 	"k8s.io/client-go/kubernetes/scheme"
 
+	"github.com/flant/shell-operator/pkg/utils/manifest"
```
zuzzas, line 15: group with other flant/shell-operator statements

pkg/kube/fake/cluster.go
```diff
@@ -156,3 +176,28 @@ func findGvr(resources []*metav1.APIResourceList, apiVersion, kindOrName string)
 	}
 	return nil
 }
+
+// Pluralize simplest way to make plural form (like resource) from object Kind
```
zuzzas, line 180:
```suggestion
// Pluralize() is the simplest way to make a plural form (like resource) from object Kind
```

pkg/kube/object_patch/patch.go
```diff
@@ -63,6 +106,22 @@ func (o *ObjectPatcher) GenerateFromJSONAndExecuteOperations(specs []OperationSp
 	for _, spec := range specs {
 		log.Debugf("Applying spec: %s", spew.Sdump(spec))
 
+		switch spec.Operation {
+		case Create, CreateIfNotExists, CreateOrUpdate, MergePatch, JSONPatch:
+		// pass - we dont need it for operations with Object
+
+		default:
+			// preferred resources does not exist on FakeCluster
+			// try to iterate and found first match
```
zuzzas, line 115:
```suggestion
			// try to iterate and find the first match
```

pkg/kube/object_patch/patch.go
```diff
@@ -63,6 +106,22 @@ func (o *ObjectPatcher) GenerateFromJSONAndExecuteOperations(specs []OperationSp
 	for _, spec := range specs {
 		log.Debugf("Applying spec: %s", spew.Sdump(spec))
 
+		switch spec.Operation {
+		case Create, CreateIfNotExists, CreateOrUpdate, MergePatch, JSONPatch:
+		// pass - we dont need it for operations with Object
+
+		default:
+			// preferred resources does not exist on FakeCluster
```
zuzzas, line 114:
```suggestion
			// preferred resources does not exist in the FakeCluster
```

pkg/kube/object_patch/patch.go
```diff
@@ -63,6 +106,22 @@ func (o *ObjectPatcher) GenerateFromJSONAndExecuteOperations(specs []OperationSp
 	for _, spec := range specs {
 		log.Debugf("Applying spec: %s", spew.Sdump(spec))
 
+		switch spec.Operation {
+		case Create, CreateIfNotExists, CreateOrUpdate, MergePatch, JSONPatch:
+		// pass - we dont need it for operations with Object
```
zuzzas, line 111:
```suggestion
		// skip - we dont need it for operations with Object
```

pkg/kube/object_patch/patch.go
```diff
@@ -55,6 +65,39 @@ func ParseSpecs(specBytes []byte) ([]OperationSpec, error) {
 	return specs, validationErrors.ErrorOrNil()
 }
 
+func (o *ObjectPatcher) fakeClusterPreferredVersion(spec OperationSpec) (OperationSpec, error) {
```
zuzzas, line 68: The name of the function is misleading since it also applies to the real client. `clusterPreferredVersion`?

test/hook/context/state.go
```diff
@@ -8,7 +8,9 @@ import (
 	"github.com/flant/shell-operator/pkg/utils/manifest"
 )
 
-var defaultNamespace = "default"
+// if we use default, then we are not able to emulate global resources due to fake cluster limitations
+// that's why empty namespace, specify default by your own if needed
```
zuzzas, line 12:
```suggestion
// that's why we use an empty namespace, specify default on your own if needed
```

zuzzas (changes requested):

pkg/kube/object_patch/patch.go
yalosev: Well, it's not applied to the real client because of
```
if len(preferred) > 0 {
	// its not a fake cluster, we can hold empty version, it works for production
	return spec, nil
}
```

inside the function

pkg/kube/fake/cluster.go
yalosev: done

pkg/kube/fake/cluster.go
yalosev: fixed almost like that

test/hook/context/state.go
yalosev: fixed

pkg/kube/object_patch/patch.go
zuzzas: Nice.

zuzzas (approved):

nabokihms (approved):
Looks good, and e2e passed.

pkg/kube/client.go
```diff
@@ -65,7 +66,7 @@ var NewKubernetesClient = func() KubernetesClient {
 	return &kubernetesClient{}
 }
 
-func NewFakeKubernetesClient() KubernetesClient {
+func NewFakeKubernetesClient(_ map[schema.GroupVersionResource]string) KubernetesClient {
```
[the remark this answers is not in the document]
yalosev: It has commentary inside.  It was placed here for 0.20.x client but at the moment I have no idea how to work with CRD in 0.20.x, thats why this code was rolled back. Anyway, someday we will have to migrate further.

PR #295, opened 2021-06-04
feat: add and remove kubernetes bindings in hook controller

diafour:
#### Overview

Implement UpdateMonitor method to change kind and apiVersion of an existing monitor.

This PR is a base for PR in addon-operator https://github.com/flant/addon-operator/pull/216

```release-note

```

### DIFF
--- a/pkg/hook/controller/kubernetes_bindings_controller.go
+++ b/pkg/hook/controller/kubernetes_bindings_controller.go
@@ -7,9 +7,9 @@ import (
 
 	. "github.com/flant/shell-operator/pkg/hook/binding_context"
 	. "github.com/flant/shell-operator/pkg/hook/types"
-	. "github.com/flant/shell-operator/pkg/kube_events_manager/types"
-
 	"github.com/flant/shell-operator/pkg/kube_events_manager"
+	. "github.com/flant/shell-operator/pkg/kube_events_manager/types"
+	utils "github.com/flant/shell-operator/pkg/utils/labels"
 )
 
 // A link between a binding config and a Monitor.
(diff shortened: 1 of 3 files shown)


pkg/hook/controller/kubernetes_bindings_controller.go
```diff
@@ -99,6 +100,57 @@ func (c *kubernetesBindingsController) EnableKubernetesBindings() ([]BindingExec
 	return res, nil
 }
 
+func (c *kubernetesBindingsController) UpdateMonitor(monitorId string, kind, apiVersion string) error {
```
nabokihms, line 103: We will face the problem of executing this method concurrently, won't we?

pkg/hook/controller/kubernetes_bindings_controller.go
```diff
@@ -99,6 +100,57 @@ func (c *kubernetesBindingsController) EnableKubernetesBindings() ([]BindingExec
...
+	if err != nil {
+		return fmt.Errorf("stop monitor for binding '%s': %v", bindingName, err)
+	}
+
+	// remove snapshots cache for monitor
+	// TODO is it required?
+	delete(c.snapshotsCache, bindingName)
+
+	// Monitor config is a pointer, so is should be updated.
+	if link.BindingConfig.Monitor.Kind != kind {
```
nabokihms, line 122: I think we should also check the API version here.

pkg/hook/controller/kubernetes_bindings_controller.go
```diff
@@ -99,6 +100,57 @@ func (c *kubernetesBindingsController) EnableKubernetesBindings() ([]BindingExec
...
+		log.Infof("Monitor in hook controller is not updated: kind='%s', apiVersion='%s'. Desired: kind='%s', apiVersion='%s'. I force update them now.",
+			link.BindingConfig.Monitor.Kind, link.BindingConfig.Monitor.ApiVersion,
```
nabokihms, line 124: "I force them to update now" sounds frightening a little :) Could we rewrite it more formally?

pkg/hook/controller/kubernetes_bindings_controller.go
diafour: I thought about this situation. It can happen for the hook with different queues but Go hook has only one queue, so there is no way to execute UpdateMonitor with one `monitorId` in parallel.

pkg/hook/controller/kubernetes_bindings_controller.go
diafour: Done.

pkg/hook/controller/kubernetes_bindings_controller.go
diafour: I've rewritten and moved this log message.

pkg/hook/controller/kubernetes_bindings_controller.go
nabokihms: Make sense. It will be a problem if we make shared monitors, but I agree that for now, everything is ok.

nabokihms (approved):

PR #298, opened 2021-06-16
Add SpecGenerator which is compatible with ObjectPatcher

yalosev:
#### Overview

Add SpecGenerator to create SpecOperations for ObjectPatcher instead on instantly executing

Used to collect spec in go-hooks and then execute them at once
after hook has been completed. We need it for making the same behavior for shell-hooks and go-hooks 

NONE

```release-note

```

### DIFF
--- a/pkg/kube/object_patch/spec_generator.go
+++ b/pkg/kube/object_patch/spec_generator.go
@@ -1,0 +1,156 @@
+package object_patch
+
+import (
+	"encoding/json"
+	"fmt"
+
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+)
+
+type specGenerator struct {
+	patchOperations []OperationSpec
+}
+
+// NewSpecGenerator creates OperationSpec generator which is compatible with ObjectPatcher interface
+func NewSpecGenerator() *specGenerator {
+	return &specGenerator{
+		patchOperations: make([]OperationSpec, 0),
+	}
+}
+
+func (dop *specGenerator) CreateOrUpdateObject(object *unstructured.Unstructured, subresource string) error {
+	return dop.create(CreateOrUpdate, object, subresource)
+}
+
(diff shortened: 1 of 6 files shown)


diafour (approved):

pkg/metric_storage/metric_storage.go
```diff
@@ -374,6 +375,7 @@ func (m *MetricStorage) ApplyOperation(op operation.MetricOperation, commonLabel
 		m.GaugeSet(op.Name, *op.Value, labels)
 		return
 	}
+	// nolint:staticcheck
```
nabokihms, line 378: Why do we need to ignore the staticcheck here? Maybe we can add this case to the global exclude settings instead?

nabokihms (approved):
LGTM besides a small question.

pkg/kube/object_patch/spec_generator.go
```diff
@@ -0,0 +1,156 @@
+package object_patch
+
+import (
+	"encoding/json"
+	"fmt"
+
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+)
+
+type specGenerator struct {
```
diafour, line 10: I think "SpecGenerator" is confusing. It is more a `PatchCollector` — similar to the `MetricCollector`.

pkg/metric_storage/metric_storage.go
yalosev: Because Add and Set properties for MetricOperation were marked as Deprecated and staticcheck linter does not allow to use them anymore. Thats why old code marked as nolint. Don't think that its a good idea to make a global exclude for this case

pkg/kube/object_patch/spec_generator.go
yalosev: Indeed. It was something like `DeferredObjectPatcher` but PatchCollector look great

diafour (approved):

PR #309, opened 2021-08-18
Fix snapshot reference in HOOKS doc

krbcmt:
#### Overview
Correctly reference the right snapshot name, _I think_

I was reading through the docs and this didn't make sense to me so I've made a change to what I expect it to be. If my change is in error could you explain what the "cm" is in reference to? Thanks.

```
Fix snapshot reference in HOOKS doc
```

### DIFF
--- a/HOOKS.md
+++ b/HOOKS.md
@@ -523,7 +523,7 @@ configVersion: v1
 schedule:
 - name: periodic-checking
   crontab: "0 */3 * * *"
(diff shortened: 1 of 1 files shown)


zuzzas:
It referred to the "ConfigMap". Your fix is absolutely correct.

zuzzas (approved):

diafour:
Nice catch! Thank you!

PR #319, opened 2021-09-15
Update Dockerfile

arkhipovds:
Just "python" doesn't work

```release-note

```


diafour:
Oh, it's the recent alpine. Thank you.

PR #349, opened 2022-01-29
Add initial delay for cron jobs

name212:
#### Overview

Add `firstRunDelay` setting for crontab hooks.
This setting add delay before first schedule hook.

In [addon-operator](https://github.com/flant/addon-operator), we want run crontab tasks after first converge.
For prevent running multiple hooks in one time, we need to add delay for first run.
We added `firstRunDelay` setting for crontab hooks.

Maybe, we can choice better name for setting?

Yes. User can add first run hook delay with `firstRunDelay` setting.
This setting is optional.

```release-note

```

### DIFF
--- a/HOOKS.md
+++ b/HOOKS.md
@@ -124,6 +124,12 @@ schedule:
   crontab: "* * * * *"
   allowFailure: true|false
   group: "pods"
+
+- name: "every minute with first run delay"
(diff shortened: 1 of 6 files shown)


name212:
Close, because we decide to decline from this feature

PR #383, opened 2022-04-12
Add support from proxying JSON logs from hooks

dulvac:
#### Overview
This PR allows for a new configurable logging type which expects JSON in each line that is output by the hooks, enhances that JSON with the shell-operator log labels and just writes them to the writer directly, without using its internal logging mechanism.

It allows for structured logging from the hooks themselves, which can now add their own "level" field, add more top-level fields, etc; while retaining the "shell-operator" fields
We need it for structured logging support in our operator 

yes, a new logging flag

```release-note
structured logging support for hooks
```


dulvac:
@nabokihms, @diafour we have another contribution. It would be great if you could have a look at it. It should be pretty simple and seems risk-free, as it doesn't change existing behaviour.

diafour:
Hello! Thanks for your time!

I thought about this problem recently and had some consideration for log proxying:
1. Possible JSON log message is detectable by `{` at the beginning.
2. Text log is difficult to parse, enable proxy only for JSON log messages.
3. shell-operator add fields: hook, binding, event, task, queue. logrus use level, time and msg fields. Should proxy override these fields obtained from the hook?
3.1. Log level should be respected for proxied logs.
3.2. There is `--log-no-time` and I'm sure we want timestamp consistency.
3.3. There is no harm if hook sets own value for `binding` field. Why not?
3.4. So, just add "JSON logs proxying rules" to the documentation: (a) use 'level' and 'msg' fields, (b) 'time' field will be overridden or dropped, (c) try not to change 'hook', 'binding', 'event', 'task', 'queue' fields, but no harm if you do.

So:
- The new flag seems superfluous.
- `fmt.Fprintf` always print messages, no respect to the log level from the hook
- Hook can not set own values for 'hook', 'binding', 'event', 'task', 'queue' fields.
- Log messages from the hook will always have time field.

It'll be great if you:
- Replace flag with simple JSON detection.
- Construct logEntry with labels from parsed data (without time, level, msg).
- Use 'level' and 'msg' to add log message via new LogEntry.
- Add proxying rules to HOOKS.md.

dulvac:
@diafour , I'll try and tackle these one by one, because we've given this a lot of thought.
> 1. Possible JSON log message is detectable by { at the beginning.

This is too much magic imo. I think it's much better to give control to the owner of the deployment. What if hooks don't normally output json lines, but rather text and one line says something like `{{template.MYVAL}} is not set!` (random example). It would incorrectly think it's json. With the independent flag, we're sure it's one or the other

> 2. Text log is difficult to parse, enable proxy only for JSON log messages.

I'm not sure I get this point. There is a fallback mechanism in my PR in case there's an unexpected stacktrace or panic (we use go for the hooks) and the hook subprocess outputs something else than json in a line. 

>3.  shell-operator add fields: hook, binding, event, task, queue. logrus use level, time and msg fields. Should proxy override these fields obtained from the hook?

Our intention is to control the loglevel from the hook - which is not possible today, everything from the hook is `info` -, to add top-level fields like `timestamp` or _any_ other values that have structured value (we use zerolog in the hooks, but it doesn't matter).

In this PR, ONLY _hook_, _binding_, _event_, _task_, _queue_ are potentially overwritten, which I think is a fair thing to constrain on the hooks, saying that these are reserved. The timestamp is a debatable one, but it is internal to the logrus logger (same as "msg"), which is not exercised at all when doing the proxying. We chose to use the same field and format from the hook logger. But the overall idea is to have the option (read "flag") to give a lot of control to the hooks, if needed.

> fmt.Fprintf always print messages, no respect to the log level from the hook

One of the elements of control that we want is the ability to control from the hook what gets logged or not. If we print it to stdout/ stderr with a "debug" level, it means we want to print debug messages (this is a separate config or it can make use of the same env var as the shell-operator parent process). Like this, we can even disable debug logs from shell-operator and enable debug messages from hooks (even independently, per hook)

> Construct logEntry with labels from parsed data (without time, level, msg).

I hope I can convince you that a hook that dumps json usually uses a logger of some kind and matching the feature sets and reconciling control between the shell-operator logging mechanism and the hook logger would be difficult, risky and frankly unnecessary. tl;dr; we just want to do structured logging from the hook directly. Having the internal fields like _hook_ and _binding_ is nice, but that's not the main goal. 

Other options include a workaround of writing to a file from hooks instead of stdout, but we kind of want to keep it 12-factor and keep all logging to stdout/ stderr

> Add proxying rules to HOOKS.md.

Will do!

diafour:
> What if hooks don't normally output json lines, but rather text and one line says something like {{template.MYVAL}} is not set! (random example). It would incorrectly think it's json. With the independent flag, we're sure it's one or the other

Your changes have a fallback to print as-is if Unmarshal fails for these messages. But I agree about "too much magic". Let's stick with adding `--log-proxy-hook-json`.

> 2. Text log is difficult to parse, enable proxy only for JSON log messages.

I mean that hook author can use structural logging with fields and enable text format (e.g. logrus can do it). Logs proxy works only with JSON output from hook. So it is more an another notice in documentation.

> Our intention is to control the loglevel from the hook
> ...
> it can make use of the same env var as the shell-operator parent process
> ...
> Like this, we can even disable debug logs from shell-operator and enable debug messages from hooks (even independently, per hook)

We've got a new feature to change loglevel in runtime. I think 'hook executor' will be able to get loglevel from runtime config and pass effective LOG_LEVEL to the hook. (It's an idea for another PR).

> Construct logEntry with labels from parsed data (without time, level, msg).

I propose to use new LogEntry instead of fmt.Fprintf because of these thoughts:
1. Force "global" log level.
2. Force "--log-no-time"
3. Transform JSON message from hook to text when `LOG_TYPE=text` is set.
4. Prevent overlaps in printing.

You've gave me arguments against 1 and 2. Point 3 seems another notice in documentation, e.g. "Using `--log-proxy-hook-json` with `--log-type=text|collor` lead to inconsistent log format".

The last point still with us: Fprintf and underlying write calls are not safe. logrus uses locking to achieve consistent output, but there is no way to write custom buffer using it's lock. zerolog defines a [`syncWriter`](https://github.com/rs/zerolog/blob/master/writer.go#L28-L56) wrapper to achieve consistency. It seems we need to wrap Logger.Out with something like this if `--log-proxy-hook-json` is enabled.

In summary, now I'm OK with your solution, only documentation and Fprintf bothers me.

dulvac:
@diafour , sorry for the late reply; I was on holiday.
Thanks for the clarification. I agree the last point standing is problematic. What you're saying makes a lot of sense. I now need to figure out how to use `LogEntry` and not have logrus e.g. wrap the (already-json) log message into its own json or add more "text" fields (depending on the configured logging mode). Do you have any tips?

pkg/app/log.go
```diff
@@ -76,3 +88,18 @@ func setLogLevel(logLevel string) {
...
+
+type ProxyJsonWrapperFormatter struct {
+	WrappedFormatter log.Formatter
+}
+
+func (f *ProxyJsonWrapperFormatter) Format(entry *log.Entry) ([]byte, error) {
+	// if proxying the json message is intended, just return the bytes
+	// TODO: Find a more elegant way to carry this info
+	if entry.Data[ProxyJsonLogKey] == true {
+		return []byte(entry.Message), nil
```
diafour, line 100:
```suggestion
		b := bytes.NewBufferString(entry.Message)
		b.WriteString("\n")
		return b.Bytes(), nil
```

I've tried this new version and log lines are glued together because `json.Marshal` returns buffer without `\n`. The original JSONFormatter uses `NewEncoder.Encode` which adds `\n`, but adding the newline here is more simpler solution.

diafour:
It looks OK! I can merge this PR if you OK with the suggestion about the newline.

Some notes:
- I've tried to use SyncWriter and got a deadlock (`fatal error: all goroutines are asleep - deadlock!`).
- Another solution is to use log.WithContext and pass a context with value.
- It'll be good to create logEntry with the field at the beginning of `proxyJSONLogs` to slightly reduce allocations per log line, but compared to Marshal/Unmarshal its not a big deal.

dulvac:
Thanks @diafour!

I appreciate the involvement. If you can apply the suggestion and merge, that would be great. I'll look into your other points and most probably keep contributing.

Thanks!

diafour (approved):

dulvac:
Thanks a lot @diafour

PR #388, opened 2022-04-26
feat: 'keep' status for tasks, summary in queue dump

diafour:
#### Overview

Changes to support 'multiphase tasks' in addon-operator:
- 'keep' status for tasks
- new method to cancel task delay

Also:
- bootstrap refactoring
- summary in queue dump

This is a supporting PR for https://github.com/flant/addon-operator/pull/315. 

```release-note

```


yalosev (approved):

pkg/shell-operator/bootstrap.go
```diff
@@ -0,0 +1,237 @@
...
+	op.KubeEventsManager.WithKubeClient(op.KubeClient)
+	op.KubeEventsManager.WithContext(op.ctx)
+	op.KubeEventsManager.WithMetricStorage(op.MetricStorage)
+
+	// Initialize events handler that emit tasks to run hooks
+	op.ManagerEventsHandler = NewManagerEventsHandler()
+	op.ManagerEventsHandler.WithContext(op.ctx)
+	op.ManagerEventsHandler.WithTaskQueueSet(op.TaskQueues)
+	op.ManagerEventsHandler.WithScheduleManager(op.ScheduleManager)
+	op.ManagerEventsHandler.WithKubeEventsManager(op.KubeEventsManager)
```
yalosev, line 213: Most of these `WithXXX` looks weird honestly :( They are required for workability but you can miss this call. Don't think we should do it at the moment but it could be a point in future. To make it as required input args for constructor

pkg/shell-operator/debug_server.go
```diff
@@ -0,0 +1,97 @@
...
+	"github.com/flant/shell-operator/pkg/app"
+	"github.com/flant/shell-operator/pkg/config"
+	"github.com/flant/shell-operator/pkg/debug"
+	"github.com/flant/shell-operator/pkg/task/dump"
+)
+
+func DefaultDebugServer() *debug.Server {
+	dbgSrv := debug.NewServer()
+	dbgSrv.WithPrefix("/debug")
+	dbgSrv.WithSocketPath(app.DebugUnixSocket)
```
yalosev, line 19: You have changed it [here](https://github.com/flant/shell-operator/pull/388/files#diff-9390d7fe36fa084015519d7717a8b8b3d1ae1c3c73f888705ee3903fef61793cR40)
But left `DebugUnixSocket` here. Is it ok?

pkg/shell-operator/bootstrap.go
diafour: Yes, one day we'll refactor these "conventions" as well. This PR only takes out initialization responsibility from ShellOperator's methods.

yalosev (approved):

PR #407, opened 2022-06-15
fix: restore exponential backoff after #388

diafour:
#### Overview

Restore exponential backoff for delayed between  failed hooks restarts.

Failed hook restarts every second after merging #388.

```release-note

```

### DIFF
--- a/pkg/task/queue/task_queue.go
+++ b/pkg/task/queue/task_queue.go
@@ -28,10 +28,10 @@ config parameter.
 */
 
 var (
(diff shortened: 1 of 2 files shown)


yalosev (approved):

nabokihms (changes requested):
I would like to see a test for the backoff feature because otherwise, we may lose it again during one of the future refactorings.

diafour:
Added tests for exponential delaying and for CancelTaskDelay.

yalosev (approved):

PR #399, opened 2022-05-24
Enable EnableKubeEventCb in dynamicaly created resource informer

and-1:
#### Overview

Dynamic resource informer created with disabled eventCbEnabled flag, so created resources in new namespace does not handled.

#### What this PR does / why we need it

Fix #379 issue

Issue #379, opened 2022-04-11
Shell-operator does not create task to exec when received event


and-1:
**Expected behavior**:
Shell-operator handle new event

**Actual behavior**:
Shell-operator receive event but doesn't handle it, doesn't create task for this.

**Steps to reproduce**:
1. Create hook with config like that:
<pre>
{
   "configVersion":"v1",
   "kubernetes": [
      {
        "name": "GIT-SYNC-PD",
        "allowFailure": false,
        "apiVersion": "kubeflow.org/v1alpha1",
        "kind": "PodDefault",
        "executeHookOnEvent": ["Added","Deleted"],
        "labelSelector": {
          "matchLabels": {
            "service/git-sync": "part-of",
          }
        },
        "namespace": {
          "labelSelector": {
            "matchLabels": {
              "app.kubernetes.io/part-of": "kubeflow-profile"
            }
          }
        },
        "queue": "git-sync"
      }
   ]
}
</pre>
2. Create namespace with label "app.kubernetes.io/part-of": "kubeflow-profile"
3. Create poddefault in created namespace, like that:
<pre>
apiVersion: kubeflow.org/v1alpha1
kind: PodDefault
metadata:
  labels:
    service/git-sync: part-of
  name: git-repository-access
  namespace: user-namespace
spec:
  desc: Access to GIT repository
  labels:
    label/added: "test"
</pre>

**Environment**:
- Shell-operator version: v1.0.9
- Kubernetes version: 1.18.10
- Installation type: helm

<details><summary>Logs</summary><br><pre>
{"level":"debug","msg":"kubernetes[1]{GIT-SYNC-PD}: Added user-namespace/PodDefault/git-repository-access: send KubeEvent","time":"2022-04-11T18:33:32+03:00"}
</pre></details>

In pod log i see only line about event, nothing else (Usual after such line followed line like "Create tasks for 'kubernetes' event 'Event 'Added' for ...". If i recreate(delete/apply) poddefault in namespace created before started shell-operator - hook executed successful. If i restart shell-operator and recreate(delete/apply) poddefault resource in "user-namespace" namespace - hook executed successful.

diafour:
Hello! Thank you for report. Seems a bug in "dynamic namespace" support. I'll check it.

and-1:
any updates?

nabokihms:
Should be fixed by https://github.com/flant/shell-operator/pull/411
I'm going to close this. If you still have a problem, feel free to open a new issue.

PR #411, opened 2022-07-12
fix: enable kube events for hooks with namespace.labelSelector bindings

diafour:
#### Overview

Add flag to enable KubeEvents for varying informers after Synchronization phase.

See #379 and #399. This fix enables events for hooks with namespace.labelSelector.

"before" situation can be emulated by commenting `m.eventsEnabled = true` in monitor.go. Test will fail:

```
    monitor_test.go:119: 
        Timed out after 5.000s.
        Should fire KubeEvent after Synchronization
        Expected
            <bool>: false
        to be true
```

```release-note

```

### DIFF
--- a/pkg/kube_events_manager/monitor.go
+++ b/pkg/kube_events_manager/monitor.go
@@ -45,7 +45,8 @@ type monitor struct {
 
 	informerSyncTime time.Duration
 
-	eventCb func(KubeEvent)
+	eventCb       func(KubeEvent)
+	eventsEnabled bool
 	// Index of namespaces statically defined in monitor configuration
 	staticNamespaces map[string]bool
 
@@ -135,7 +136,7 @@ func (m *monitor) CreateInformers() error {
 		m.NamespaceInformer.WithKubeClient(m.KubeClient)
 		err := m.NamespaceInformer.CreateSharedInformer(
 			func(nsName string) {
-				// add function — check, create and run informers for Ns
+				// Added/Modified event: check, create and run informers for Ns
 				// ignore event if namespace is already has static ResourceInformers
 				if _, ok := m.staticNamespaces[nsName]; ok {
 					return
@@ -160,11 +161,13 @@ func (m *monitor) CreateInformers() error {
 				for _, informer := range m.VaryingInformers[nsName] {
 					informer.WithContext(ctx)
 					informer.Start()
-					informer.EnableKubeEventCb()
+					if m.eventsEnabled {
+						informer.EnableKubeEventCb()
+					}
 				}
 			},
 			func(nsName string) {
-				// delete function — check, stop and remove informers for Ns
+				// Delete event: check, stop and remove informers for Ns
 				logEntry.Infof("deleted ns/%s, stop dynamic ResourceInformers", nsName)
 
 				// ignore statically specified namespaces
@@ -239,6 +242,8 @@ func (m *monitor) EnableKubeEventCb() {
 			informer.EnableKubeEventCb()
 		}
 	}
+	// Enable events for future VaryingInformers.
+	m.eventsEnabled = true
 }
 
 // CreateInformersForNamespace creates informers bounded to the namespace. If no matchName is specified,
--- a/pkg/kube_events_manager/monitor.go
+++ b/pkg/kube_events_manager/monitor.go
@@ -160,10 +160,10 @@ func (m *monitor) CreateInformers() error {
 
 				for _, informer := range m.VaryingInformers[nsName] {
 					informer.WithContext(ctx)
-					informer.Start()
 					if m.eventsEnabled {
 						informer.EnableKubeEventCb()
 					}
+					informer.Start()
 				}
 			},
 			func(nsName string) {
--- a/pkg/kube_events_manager/monitor.go
+++ b/pkg/kube_events_manager/monitor.go
@@ -136,8 +136,6 @@ func (m *monitor) CreateInformers() error {
 		err := m.NamespaceInformer.CreateSharedInformer(
 			func(nsName string) {
 				// add function — check, create and run informers for Ns
-				logEntry.Infof("got ns/%s, create dynamic ResourceInformers", nsName)
-
 				// ignore event if namespace is already has static ResourceInformers
 				if _, ok := m.staticNamespaces[nsName]; ok {
 					return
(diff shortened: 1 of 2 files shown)


pkg/kube_events_manager/monitor_test.go
```diff
@@ -0,0 +1,143 @@
...
+	}, "5s", "10ms").Should(BeTrue(), "Should create informer for new Namespace")
+
+	createCM(fc, "test-ns-1", testCM("cm-1"))
+
+	// Should update snapshot. Snapshot resets cached events.
+	g.Eventually(func() bool {
+		snap := mon.Snapshot()
+		return len(snap) == 2
+	}, "5s", "10ms").Should(BeTrue(), "Should not update snapshot until EnableKubeEventsCb")
+
```
shvgn, line 83: The comment says "should update", while assertion description says "should not"

pkg/kube_events_manager/monitor_test.go
```diff
@@ -0,0 +1,143 @@
...
+	// Create more ConfigMaps to cache some events.
+	createCM(fc, "test-ns-1", testCM("cm-2"))
+	createCM(fc, "test-ns-1", testCM("cm-3"))
+
+	g.Expect(events).Should(HaveLen(0), "Should not fire KubeEvents until EnableKubeEventsCb")
+
+	// Enable Kube events, should update snapshots now.
+	mon.EnableKubeEventCb()
```
shvgn, line 91: Should we check `len(events) == 0` before this call?

pkg/kube_events_manager/monitor_test.go
```diff
@@ -0,0 +1,143 @@
...
+
+	// Now simulate NS creation and objects creation after Synchronization phase.
+
+	// Create new ns with labels and cm there.
+	createNsWithLabels(fc, "test-ns-2", map[string]string{"test-label": ""})
+
+	// Wait until informer for new NS appears.
+	g.Eventually(func() bool {
+		monImpl := mon.(*monitor)
+		return len(monImpl.VaryingInformers) == 3
```
shvgn, line 106: Counting seems good enough since this behavior looks deterministic. But I suggest to be more strict and test that `VaryingInformers` actually contains key "test-ns-2"

pkg/kube_events_manager/monitor_test.go
```diff
@@ -0,0 +1,143 @@
...
+	g.Expect(events).Should(HaveLen(0), "Should not call KubeEventCb until EnableKubeEventsCb")
+
+	// Simulate object creation during Synchronization phase:
+	// create new ns with matching labels and then create new ConfigMap.
+	createNsWithLabels(fc, "test-ns-1", map[string]string{"test-label": ""})
+
+	// Wait until informer appears.
+	g.Eventually(func() bool {
+		monImpl := mon.(*monitor)
+		return len(monImpl.VaryingInformers) == 2
```
shvgn, line 73: Counting seems good enough since this behavior looks deterministic. But I suggest to be more strict and test that `VaryingInformers` actually contains expected keys

pkg/kube_events_manager/monitor_test.go
```diff
@@ -0,0 +1,143 @@
...
+	g.Eventually(func() bool {
+		monImpl := mon.(*monitor)
+		return len(monImpl.VaryingInformers) == 3
+	}, "5s", "10ms").Should(BeTrue(), "Should create informer for ns/test-ns-2")
+
+	createCM(fc, "test-ns-2", testCM("cm-2-1"))
+
+	// Should update snapshot.
+	g.Eventually(func() bool {
+		return len(mon.Snapshot()) == 4
```
shvgn, line 113: I would prefer to find exact object reference in snapshots. We probably don't case about the amount, but we do care about the inclusion of the configmap. What if there is a bug that copies `snap[0]` all the time orgrows the slice with nils? :trollface:

pkg/kube_events_manager/monitor.go
```diff
@@ -160,10 +161,13 @@ func (m *monitor) CreateInformers() error {
 				for _, informer := range m.VaryingInformers[nsName] {
 					informer.WithContext(ctx)
 					informer.Start()
+					if m.eventsEnabled {
+						informer.EnableKubeEventCb()
+					}
```
shvgn, line 166: For me it is not too much clear, why `CreateInformers` can be called multiple times. 

Probably there could be single config on the start, and it should be called only once per application lifecycle. Seems that is not the case.

Then, since this method is public, I thought it could be called by clients. Bu I think it is only called on the addition of configs when hooks are initialized. And the call is hidden under Monitor interface. That became cleaner to me.

What if `CreateInformers` was private? Can we make it private, assuming nobody commited to use this interface?

shvgn (changes requested):
I'd prefer strict assertions.

pkg/kube_events_manager/monitor.go
diafour: `CreateInformers` should be called once, it is more an `Init` method that configure monitor and may return error. If this initialization was successful, monitor can be `Start`ed and `Stop`ed multiple times.

`informer.EnableKubeEventCb()` is used inside an Add/Modify callback for the namespace informer. So this block can be called multiply times on changes in namespace labels.

pkg/kube_events_manager/monitor_test.go
diafour: I've rewrite this test.

pkg/kube_events_manager/monitor_test.go
diafour: Done

yalosev (approved):

pkg/kube_events_manager/monitor.go
diafour: `monitor.EnableKubeEventCb()` is called after successful hook execution with Synchronization context.

`monitor.CreateInformers` creates an informer for Namespaces if NamespaceSelector.LabelSelector is defined in MonitorConfig. This informer has two callbacks for Add/Modify and Delete events from namespaces. So, `monitor.eventsEnabled` will be checked every time when new namespace added or deleted, not once during `monitor.CreateInformers` call.

shvgn (approved):

nabokihms (approved):
Lgtm

PR #412, opened 2022-07-14
Use Factory for informers

nabokihms:
#### Overview

Reuse watch and cache for informers

The most resources shell-operator consume to encode and decode watch events.

```release-note

```

### DIFF
--- a/pkg/kube_events_manager/monitor.go
+++ b/pkg/kube_events_manager/monitor.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"fmt"
 	"sort"
-	"time"
 
 	log "github.com/sirupsen/logrus"
 
@@ -20,7 +19,6 @@ type Monitor interface {
 	WithMetricStorage(mstor *metric_storage.MetricStorage)
 	WithConfig(config *MonitorConfig)
 	WithKubeEventCb(eventCb func(KubeEvent))
-	WithSyncPeriod(time.Duration)
 	CreateInformers() error
 	Start(context.Context)
 	Stop()
@@ -43,8 +41,6 @@ type monitor struct {
 	// map of dynamically starting informers
 	VaryingInformers map[string][]ResourceInformer
 
-	informerSyncTime time.Duration
-
 	eventCb       func(KubeEvent)
 	eventsEnabled bool
 	// Index of namespaces statically defined in monitor configuration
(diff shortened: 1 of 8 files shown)


pkg/kube_events_manager/monitor.go
```diff
@@ -299,7 +288,6 @@ func (m *monitor) Start(parentCtx context.Context) {
 
 	if m.NamespaceInformer != nil {
 		m.NamespaceInformer.WithContext(m.ctx)
-		m.NamespaceInformer.WithSyncPeriod(m.informerSyncTime)
```
diafour, line 302: NamespaceInformer is not using factory. Is it OK to delete this setter and use 0 as syncTime in underlying `wait.PollImmediateUntil`?

nabokihms: Good question. I used `WithSyncPeriod` previously to tune this for tests to reduce the startup time for tests, Since we start this informer only once, tuning the sync period doesn't win us much time and look like a complication without benefits.

pkg/kube_events_manager/resource_informer.go
```diff
@@ -479,29 +461,26 @@ func (ei *resourceInformer) ShouldFireEvent(checkEvent WatchEventType) bool {
...
 func (ei *resourceInformer) Start() {
 	log.Debugf("%s: RUN resource informer", ei.Monitor.Metadata.DebugName)
-	stopCh := make(chan struct{}, 1)
+
 	go func() {
 		<-ei.ctx.Done()
 		ei.stopped = true
-		close(stopCh)
+		DefaultFactoryStore.Stop(ei.FactoryIndex)
 	}()
```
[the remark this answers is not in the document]
nabokihms: I thought that this is still required 🙈 Could you please point me how to rearrange the code properly?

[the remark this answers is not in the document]
diafour: I think we can just delete this go routine.

pkg/kube_events_manager/monitor.go
diafour: It seems we can use DefaultSyncPeriod from factory in namespace_informer, see this commit: https://github.com/flant/shell-operator/commit/33f3aa9f45753b84f4d1586a6601d328ceaeeb4b
WDYT?

nabokihms: Not necessarily, but ok. We can do it.

diafour (approved):

PR #471, opened 2023-03-29
fix: MutatingWebhook - Snapshots not being added

abalaven:
#### Overview

When we specify values for `includeSnapshotsFrom:` and `group:` in the `KubernetesMutating` binding, the snapshots from the other bindings referenced are not added to the binding-context of `"type": "Mutating"`. The `snapshots: {}` field is empty.

This PR resolves the issue above. The snapshots accumulated from the `group:` and `includeSnapshotsFrom:` keys are added to the mutatingConfig. This is also propagated to the `KubernetesMutating` binding and the binding-context.

```release-note

```


abalaven:
Hello, just following up on this. Can any contributors please review? Thank you!

yalosev (approved):

PR #520, opened 2023-08-17
[deckhouse] Start HTTP API debug server if "DebugHttpPort" is set.

Dmitrykob:
#### Overview

Deckhouse has an API that provides info on the state of its internal storage, queues, preflight patches, and more.
This API is now exposed via the Unix socket and can be accessed with the deckhouse-controller CLI.

Added an HTTP API server on localhost next to the socket server if "DebugHttpPort" is set.

We want to make it public and secure in deckhouse using kube-rbac-proxy.

### DIFF
--- a/pkg/debug/server.go
+++ b/pkg/debug/server.go
@@ -76,7 +76,6 @@ func (s *Server) Init() (err error) {
 	}()
 
 	if s.HttpAddr != "" {
-
 		go func() {
 			if err := http.ListenAndServe(s.HttpAddr, s.Router); err != nil {
 				log.Errorf("Error starting Debug HTTP server: %s", err)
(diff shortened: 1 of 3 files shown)


pkg/debug/server.go
```diff
@@ -36,6 +37,10 @@ func (s *Server) WithPrefix(prefix string) {
 	s.Prefix = prefix
 }
 
+func (s *Server) WithHttpPort(port string) {
+	s.HttpPort = port
+}
```
yalosev, line 42: Could we refactor this a bit, please.
We don't use override or something like this in the runtime. So we can use 
```go
func NewServer(prefix, socketPath, httpPort string) *Server { }
```
signature and initialize it in the go-way

yalosev (changes requested):

yalosev (approved):

PR #524, opened 2023-09-03
improve(docs): using multiple jqFilter's

a0s:


diafour (approved):
Thank you for the improvement.

PR #538, opened 2023-10-13
add IgnoreHookError patch option

miklezzzz:
#### Overview
This PR adds additional patch_option for Filter/Patch operations named `IgnoreHookError` and some execution logic around it.

If `IgnoreHookError` is set to true for a Filter/Patch operation and `WithSubresource` equals to `/status`, such an operation will be applied even in case of the hook failure.
It can come in handy in a situation when a hook operates around some kubernetes resource and one wants to make it more obvious when the resource was noticied/processed/etc by the hook last time by means of status fields.

This way we could provide some meaningful statuses or a feedback loop on what  happens when a hook is executed.

### DIFF
--- a/pkg/kube/object_patch/operation.go
+++ b/pkg/kube/object_patch/operation.go
@@ -57,7 +57,6 @@ func GetPatchStatusOperationsOnHookError(operations []Operation) []Operation {
 			if operation.subresource == "/status" && operation.ignoreHookError {
 				patchStatusOperations = append(patchStatusOperations, operation)
 			}
-		default:
 		}
 	}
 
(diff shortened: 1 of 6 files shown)


yalosev (approved):

pkg/kube/object_patch/operation.go
```diff
@@ -43,6 +44,26 @@ const (
...
+		switch operation := op.(type) {
+		case *filterOperation:
+			if operation.subresource == "/status" && operation.ignoreHookError {
+				patchStatusOperations = append(patchStatusOperations, operation)
+			}
+		case *patchOperation:
+			if operation.subresource == "/status" && operation.ignoreHookError {
+				patchStatusOperations = append(patchStatusOperations, operation)
+			}
+		default:
```
yalosev, line 60:
```suggestion
```

PR #578, opened 2024-01-14
Update apiVersions in the crd-simple example

Alrem:
#### Overview

Update apiVersions in the crd-simple example

Kubernetes 1.27 uses updated API versions of the following kinds:
- CustomResourceDefinition
- ClusterRole
- ClusterRoleBinding

Also, CustomResourceDefinition v1 has the updated schema.


nabokihms (approved):
@Alrem thanks, good catch!

PR #609, opened 2024-04-09
Added retry budget and some refactoring.

Dmitrykob:
#### Overview

The changes are described and tested in the context of [deckhouse PR](https://github.com/deckhouse/deckhouse/pull/7949)

Add repeated queries to get CRD and apply conversion strategies.

The occasion with the use of watch has been tried. It doesn't work. Because for a successful subscription - CRD must also already be installed in the cluster.

### DIFF
--- a/pkg/webhook/conversion/crd_client_config.go
+++ b/pkg/webhook/conversion/crd_client_config.go
@@ -25,7 +25,7 @@ var SupportedConversionReviewVersions = []string{"v1", "v1beta1"}
 func (c *CrdClientConfig) Update(ctx context.Context) error {
 	var (
 		retryTimeout = 15 * time.Second
-		retryBudget  = 12 // 12 times * 15 sec = 3 min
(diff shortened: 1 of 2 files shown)


pkg/webhook/conversion/crd_client_config.go
```diff
@@ -23,43 +22,34 @@ type CrdClientConfig struct {
 
 var SupportedConversionReviewVersions = []string{"v1", "v1beta1"}
 
-func (c *CrdClientConfig) Update() error {
+func (c *CrdClientConfig) Update(ctx context.Context) error {
 	var (
 		retryTimeout = 15 * time.Second
+		retryBudget  = 12 // 12 times * 15 sec = 3 min
```
yalosev, line 28: I think we can increase it up to 10-15 minutes, to be ensured that all modules are up-to-date

yalosev (approved):

Issue #620, opened 2024-06-20
Ignore lib directory while processing hooks


yalosev:
Hook spams like this:
```
.../hooks/lib/utils.py' is skipped: no executable permissions, chmod +x is required to run this hook"
```

We can patch [this function](https://github.com/flant/shell-operator/blob/main/pkg/utils/file/file.go#L29) to ignore the `lib` directory with hooks

nabokihms:
Les't add a warning if there is a file with the executable permissions in the lib directory. I think we need this to prevent accidentally putting a hook in the folder where we ignore it.

The behavior also must be documented in all the docs: shell-operator, addon-operator, and deckhouse module development docs.

juev:
So what do we do?

The task was to get rid of many messages in the logs stating that files in the directory do not have permission to run. And it is suggested to simply ignore the lib directory from scanning.
In case we need to check that the lib directory contains executable files, we must scan it. And if there are such files, issue a warning.

Maybe then it's better not to ignore the lib directory from scanning. But just not display a warning that the file is not executable?

In this case, our system behavior will remain the same, but there will be less spam in the logs.

We just need to remove this [warning](https://github.com/flant/shell-operator/blob/main/pkg/utils/file/file.go#L57).

yalosev:
Maybe it would be better to use something like that:
```go
if !IsFileExecutable(f) && !strings.Contains(f.Name(), "/lib/") {
    log.Warnf("File '%s' is skipped: no executable permissions, chmod +x is required to run this hook", path)
    return nil
}
```

PR #621, opened 2024-06-20
feat: Ignore lib directory while processing hooks

juev:
#### Overview

Hook spams like this:
```
.../hooks/lib/utils.py' is skipped: no executable permissions, chmod +x is required to run this hook"
```

Ignore lib directory while processing hooks. 
We also additionally check the files in the lib directory; if we find files with executable permissions, we display a warning.

Closes: #620

### DIFF
--- a/pkg/hook/hook_manager.go
+++ b/pkg/hook/hook_manager.go
@@ -103,6 +103,10 @@ func (hm *hookManager) Init() error {
 	hm.hooksInOrder = make(map[BindingType][]*Hook)
 	hm.hooksByName = make(map[string]*Hook)
 
+	if err := utils_file.RecursiveCheckLibDirectory(hm.workingDir); err != nil {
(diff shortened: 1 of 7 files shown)


pkg/hook/hook_manager.go
```diff
@@ -103,6 +103,10 @@ func (hm *hookManager) Init() error {
 	hm.hooksInOrder = make(map[BindingType][]*Hook)
 	hm.hooksByName = make(map[string]*Hook)
 
+	if err := utils_file.RecursiveCheckLibDirectory(hm.workingDir); err != nil {
```
juev, line 106: We can run both functions in separate goroutines to cut the time in half.

This will be especially noticeable on projects with a large number of files in the hooks directory.

But I don't know how common this practice is in the company. Is it worth spending time on performance?

yalosev: It's a one-time operation, so we don't need this optimization now.

yalosev (approved):

PR #628, opened 2024-07-10
update merics' labelset if needed

### LINKED ISSUE #616: Lost metrics/wrong labels when same metric name has different labels
**Expected behavior (what you expected to happen)**:

Each metric producced by a hook should have exactly same labels as reported by the script

**Actual behavior (what actually happened)**:

Looks like shell-operator expects all metrics with the same name has common set of labels. When it differs between metric
instances, wrong labels are reported and some metrics can be completely omitted.

**Steps to reproduce**:

```bash
#!/usr/bin/env bash

if [[ ${1:-} == "--config" ]] ; then
  cat <<EOF
configVersion: v1
onStartup: 1
EOF
else
jq -nc '{group:"g",name: "m1", action:"set", value: 1, labels: {a: "A"}}' >>"$METRICS_PATH"
jq -nc '{group:"g",name: "m1", action:"set", value: 2, labels: {c: "C"}}' >>"$METRICS_PATH"
jq -nc '{group:"g",name: "m1", action:"set", value: 3, labels: {a: "A", b:"B"}}' >>"$METRICS_PATH"
jq -nc '{group:"g",name: "m2", action:"set", value: 1, labels: {a: "A1"}}' >>"$METRICS_PATH"
jq -nc '{group:"g",name: "m2", action:"set", value: 2, labels: {c: "C2"}}' >>"$METRICS_PATH"
jq -nc '{group:"g",name: "m2", action:"set", value: 3, labels: {a: "A3", b:"B3"}}' >>"$METRICS_PATH"
fi
```

produces:

```
# HELP m1 m1
# TYPE m1 gauge
m1{a="",hook="metrics.sh"} 2
m1{a="A",hook="metrics.sh"} 3
# HELP m2 m2
# TYPE m2 gauge
m2{a="",hook="metrics.sh"} 2
m2{a="A1",hook="metrics.sh"} 1
m2{a="A3",hook="metrics.sh"} 3
```

**Environment**:
- Shell-operator version: v1.4.7

miklezzzz:
#### Overview

This pr update the way metrics with different label sets and the same name are stored in a collector.
If a new metric has a new label, this new label is added to the collector and all previous metrics in the collector get it with an empty string as its value.
Closes #616

### DIFF
--- a/pkg/metric_storage/vault/vault.go
+++ b/pkg/metric_storage/vault/vault.go
@@ -41,6 +41,8 @@ func (v *GroupedVault) GetOrCreateCounterCollector(name string, labelNames []str
 			return nil, fmt.Errorf("counter '%s' %v registration: %v", name, labelNames, err)
 		}
 		v.collectors[name] = collector
+	} else if !LabelNamesEqual(collector.LabelNames(), labelNames) {
+		collector.UpdateLabels(labelNames)
 	}
 	if counter, ok := collector.(*ConstCounterCollector); ok {
 		return counter, nil
@@ -58,7 +60,10 @@ func (v *GroupedVault) GetOrCreateGaugeCollector(name string, labelNames []strin
 			return nil, fmt.Errorf("gauge '%s' %v registration: %v", name, labelNames, err)
 		}
 		v.collectors[name] = collector
+	} else if !LabelNamesEqual(collector.LabelNames(), labelNames) {
+		collector.UpdateLabels(labelNames)
 	}
+
 	if gauge, ok := collector.(*ConstGaugeCollector); ok {
 		return gauge, nil
 	}
--- a/pkg/utils/labels/labels.go
+++ b/pkg/utils/labels/labels.go
@@ -53,3 +53,16 @@ func DefaultIfEmpty(m map[string]string, def map[string]string) map[string]strin
 	}
 	return m
 }
+
+func LabelNamesEqual(a, b[]string) bool {
+	if len(a) != len(b) {
+		return false
+	}
+
+	for i := range a {
+		if a[i] != b[i] {
+			return false
+		}
+	}
+	return true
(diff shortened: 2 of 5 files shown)


pkg/metric_storage/vault/vault.go
```diff
@@ -58,7 +60,10 @@ func (v *GroupedVault) GetOrCreateGaugeCollector(name string, labelNames []strin
 			return nil, fmt.Errorf("gauge '%s' %v registration: %v", name, labelNames, err)
 		}
 		v.collectors[name] = collector
+	} else if !LabelNamesEqual(collector.LabelNames(), labelNames) {
+		collector.UpdateLabels(labelNames)
```
miklezzzz, line 64: not sure how to make it less noisy, maybe it makes sense to store all the collector's labels in a map alongside the slice so that it would be faster to check if any new labels are provided

pkg/utils/labels/labels.go
```diff
@@ -53,3 +53,16 @@ func DefaultIfEmpty(m map[string]string, def map[string]string) map[string]strin
...
 	return m
 }
+
+func LabelNamesEqual(a, b[]string) bool {
+	if len(a) != len(b) {
+		return false
+	}
+
+	for i := range a {
+		if a[i] != b[i] {
```
yalosev, line 63: Are these array sorted?

miklezzzz: well, in some cases they are sorted(at the beginning). but after updating a collector with new labels, they aren't sorted anymore. Maybe it makes sense to make a collector's update a bit more elaborated and also sort its labelNames slice after adding new labels. It would require some extra work to relabel all the metrics in the collector, though...

nabokihms (approved):

yalosev (approved):

PR #655, opened 2024-09-18
[kube-event-manager] fix factory context issue

miklezzzz:
#### Overview
This pr provides extra debug and fixes yet another bug in the shared factory store related to stopping all informers if the module, that triggered creation of the informers, has been disabled.

Currently, informers created for a module inherit the module monitor's context and if the module gets disabled, all the informers get disabled too. Given that we apply deduplication techniques to the factory store, disabling informers via the module's context isn't acceptable, so dedicated (per a shared informer factory) contexts were introduced.

Also, instead of using a counter for a factory to decide wether it can be safely stopped, a map of handler registrations was added. This map allows tracking of active resource informers and their handlers, and if a resource informer isn't needed any longer, its handler is removed from the factory's informer, leaving no traces behind.

### DIFF
--- a/pkg/kube_events_manager/factory.go
+++ b/pkg/kube_events_manager/factory.go
@@ -86,7 +86,7 @@ func (c *FactoryStore) get(client dynamic.Interface, index FactoryIndex) Factory
 	return c.data[index]
 }
 
-func (c *FactoryStore) Start(informerId string, ctx context.Context, client dynamic.Interface, index FactoryIndex, handler cache.ResourceEventHandler, errorHandler *WatchErrorHandler) error {
+func (c *FactoryStore) Start(ctx context.Context, informerId string, client dynamic.Interface, index FactoryIndex, handler cache.ResourceEventHandler, errorHandler *WatchErrorHandler) error {
 	c.mu.Lock()
 	defer c.mu.Unlock()
 
--- a/pkg/kube_events_manager/factory.go
+++ b/pkg/kube_events_manager/factory.go
@@ -5,6 +5,7 @@ import (
 	"sync"
 	"time"
 
+	log "github.com/sirupsen/logrus"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/util/wait"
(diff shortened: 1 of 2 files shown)


pkg/kube_events_manager/factory.go
```diff
@@ -81,21 +82,25 @@ func (c *FactoryStore) get(ctx context.Context, client dynamic.Interface, index
 		client, resyncPeriod, index.Namespace, tweakListOptions)
 	factory.ForResource(index.GVR)
 
-	c.add(ctx, index, factory)
+	c.add(index, factory)
 	return c.data[index]
 }
 
-func (c *FactoryStore) Start(ctx context.Context, client dynamic.Interface, index FactoryIndex, handler cache.ResourceEventHandler, errorHandler *WatchErrorHandler) error {
+func (c *FactoryStore) Start(informerId string, ctx context.Context, client dynamic.Interface, index FactoryIndex, handler cache.ResourceEventHandler, errorHandler *WatchErrorHandler) error {
```
yalosev, line 89: It's good form to use ctx in the first place.
```suggestion
func (c *FactoryStore) Start(ctx context.Context, informerId string, client dynamic.Interface, index FactoryIndex, handler cache.ResourceEventHandler, errorHandler *WatchErrorHandler) error {
```

pkg/kube_events_manager/resource_informer.go
```diff
@@ -427,13 +435,13 @@ func (ei *resourceInformer) start() {
...
 			<-ei.ctx.Done()
-			DefaultFactoryStore.Stop(ei.FactoryIndex)
+			DefaultFactoryStore.Stop(ei.id, ei.FactoryIndex)
 		}
 	}()
 
 	// TODO: separate handler and informer
 	errorHandler := newWatchErrorHandler(ei.Monitor.Metadata.DebugName, ei.Monitor.Kind, ei.Monitor.Metadata.LogLabels, ei.metricStorage)
-	err := DefaultFactoryStore.Start(ei.ctx, ei.KubeClient.Dynamic(), ei.FactoryIndex, ei, errorHandler)
+	err := DefaultFactoryStore.Start(ei.id, ei.ctx, ei.KubeClient.Dynamic(), ei.FactoryIndex, ei, errorHandler)
```
yalosev, line 444:
```suggestion
	err := DefaultFactoryStore.Start(ei.ctx, ei.id, ei.KubeClient.Dynamic(), ei.FactoryIndex, ei, errorHandler)
```

yalosev (approved):

PR #703, opened 2024-12-18
[ci] add nonamedreturns linter

miklezzzz:
#### Overview
Adds nonamedreturns linter.

### DIFF
--- a/pkg/hook/config/config_v1.go
+++ b/pkg/hook/config/config_v1.go
@@ -101,22 +101,21 @@ type SettingsV1 struct {
 }
 
 // ConvertAndCheck fills non-versioned structures and run inter-field checks not covered by OpenAPI schemas.
-func (cv1 *HookConfigV1) ConvertAndCheck(c *HookConfig) (err error) {
-	c.Settings, err = cv1.CheckAndConvertSettings(cv1.Settings)
-	if err != nil {
+func (cv1 *HookConfigV1) ConvertAndCheck(c *HookConfig) error {
+	var err error
+
+	if c.Settings, err = cv1.CheckAndConvertSettings(cv1.Settings); err != nil {
 		return err
 	}
 
-	c.OnStartup, err = c.ConvertOnStartup(cv1.OnStartup)
-	if err != nil {
+	if c.OnStartup, err = c.ConvertOnStartup(cv1.OnStartup); err != nil {
 		return err
 	}
 
 	c.OnKubernetesEvents = []htypes.OnKubernetesEventConfig{}
 	for i, kubeCfg := range cv1.OnKubernetesEvent {
-		err := cv1.CheckOnKubernetesEvent(kubeCfg, fmt.Sprintf("kubernetes[%d]", i))
-		if err != nil {
-			return fmt.Errorf("invalid kubernetes config [%d]: %v", i, err)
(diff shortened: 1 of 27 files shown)


pkg/hook/hook_manager.go
```diff
@@ -201,7 +201,7 @@ func (hm *Manager) loadHook(hookPath string) (hook *Hook, err error) {
 	hook.WithTmpDir(hm.TempDir())
 
 	if hook.Config == nil {
-		return nil, fmt.Errorf("hook %q is marked as executable but doesn't contain config section", hook.Path)
+		return nil, fmt.Errorf("hook %s is marked as executable but doesn't contain config section", hook.Path)
```
juev, line 204: Why are we changing %q to %s here? Thus, the hook name will now be without quotes.

pkg/hook/config/config_v1.go
```diff
@@ -520,13 +517,12 @@ func convertMutating(cfgV1 KubernetesAdmissionConfigV1) (htypes.MutatingConfig,
...
+	var allErr error
 
 	if len(cfgV1.IncludeSnapshotsFrom) > 0 {
-		err = CheckIncludeSnapshots(kubeConfigs, cfgV1.IncludeSnapshotsFrom...)
-		if err != nil {
-			allErr = multierror.Append(allErr, fmt.Errorf("includeSnapshotsFrom is invalid: %v", err))
+		if err := CheckIncludeSnapshots(kubeConfigs, cfgV1.IncludeSnapshotsFrom...); err != nil {
+			allErr = multierror.Append(allErr, fmt.Errorf("includeSnapshotsFrom is invalid: %w", err))
 		}
 	}
```
ldmonster, line 528: If i understand it correctly, it will be little short
```suggestion
	if len(cfgV1.IncludeSnapshotsFrom) > 0 {
		if err := CheckIncludeSnapshots(kubeConfigs, cfgV1.IncludeSnapshotsFrom...); err != nil {
			return fmt.Errorf("includeSnapshotsFrom is invalid: %w", err)
		}
	}

    return nil
```

juev (approved):

pkg/shell-operator/operator.go
```diff
@@ -210,13 +208,12 @@ func (op *ShellOperator) initValidatingWebhookManager() (err error) {
...
-		return
+		return nil
 	}
 
-	err = op.AdmissionWebhookManager.Init()
+	err := op.AdmissionWebhookManager.Init()
 	if err != nil {
-		log.Errorf("ValidatingWebhookManager init: %v", err)
-		return err
+		return fmt.Errorf("ConversionWebhookManager init: %w", err)
```
ldmonster, line 216: why conversion?

pkg/hook/config/config_v1.go
ldmonster: IDE makes too strange suggestion. i mean lines 524-529

miklezzzz: yep, overlooked it

pkg/hook/hook_manager.go
miklezzzz: you're right

ldmonster (approved):

yalosev (approved):

PR #717, opened 2025-02-17
Fix concurrent map access in task queues

miklezzzz:
#### Overview

Fixes sporadic concurrent map read/write access.

Based on debugging with `-race`.

### DIFF
--- a/pkg/shell-operator/manager_events_handler.go
+++ b/pkg/shell-operator/manager_events_handler.go
@@ -88,8 +88,7 @@ func (m *ManagerEventsHandler) Start() {
 
 			m.taskQueues.DoWithLock(func(tqs *queue.TaskQueueSet) {
 				for _, resTask := range tailTasks {
-					q := tqs.GetByName(resTask.GetQueueName())
-					if q == nil {
+					if q := tqs.Queues[resTask.GetQueueName()]; q == nil {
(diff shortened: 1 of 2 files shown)


pkg/shell-operator/manager_events_handler.go
```diff
@@ -88,8 +88,7 @@ func (m *ManagerEventsHandler) Start() {
 
 			m.taskQueues.DoWithLock(func(tqs *queue.TaskQueueSet) {
 				for _, resTask := range tailTasks {
-					q := tqs.GetByName(resTask.GetQueueName())
-					if q == nil {
+					if q := tqs.Queues[resTask.GetQueueName()]; q == nil {
```
ldmonster, line 91: Why we replace GetByName() here with straight read from map?
I mean, we have a great mutex using in GetByName() now. Why do not to use it here?

miklezzzz: because GetByName itself has locking in it and in this part of code GetByName is called inside DoWithLock method which also implement locking, so they conflict with each other. Yet, I believe GetByName is rather a more general (and useful) handle, while DoWithLock seems to be a debatable piece of code :)

ldmonster (approved):

PR #735, opened 2025-03-17
[shell-operator] feat/new operator event handling mechanism

ldmonster:
#### Overview
Replace functional pattern with return values


pkg/hook/controller/hook_controller.go
```diff
@@ -190,17 +202,40 @@ func (hc *HookController) HandleConversionEvent(crdName string, request *v1.Conv
...
+	}
+}
+
+func (hc *HookController) HandleScheduleEventWithFormTasks(crontab string, createTasksFn func(BindingExecutionInfo) task.Task) []task.Task {
+	if hc.ScheduleController == nil {
+		return nil
+	}
+
+	infos := hc.ScheduleController.HandleEvent(crontab)
+	if createTasksFn == nil {
```
juev, line 226: And why do we check the input parameters only after we have called HandleEvent? If the functions have not been transferred, we do not use the result that we received earlier.

ldmonster: looks like untouched code. this is not my logic. 
previously, we have no returned values, just append to slice in closure.
now we have more clear structure with returning tasks values, if we handle it.

juev: I just pointed out what the eye caught on, the correction seems to be small.

yalosev (approved):

Issue #730, opened 2025-03-11
[dependencies] Remove libjq from dependencies


ldmonster:
We need to remove libjq 'github.com/flant/libjq-go' from dependencies and add new logic with go jq library. 
For example [this](https://github.com/itchyny/gojq).

We must cover all possibly code with tests before replacing.

**Is your feature request related to a problem? Please describe.**

Shell-operator must be compiled with CGO, otherwise a jq binary file must be present next to it for proper operation.

**Describe the solution you'd like to see**

**Describe alternatives you've considered**

**Additional context**

PR #743, opened 2025-03-27
refactor: simplify jq filter initialization and remove unused libpath

juev:
#### Overview

fixed: #730

replace libjq library with itchyny/gojq


ldmonster:
We need to cover jq filter by tests

juev:
We already have [this](https://github.com/flant/shell-operator/blob/7ceb176673e8d35eb37cbfbe8c30de79ecda00f5/pkg/kube/object_patch/patch_test.go#L422-L422) and [this](https://github.com/flant/shell-operator/blob/7ceb176673e8d35eb37cbfbe8c30de79ecda00f5/pkg/kube/object_patch/patch_test.go#L441-L441) and [this](https://github.com/flant/shell-operator/blob/7ceb176673e8d35eb37cbfbe8c30de79ecda00f5/pkg/kube/object_patch/patch_test.go#L460-L460)

pkg/filter/jq/apply_test.go
```diff
@@ -0,0 +1,74 @@
+package jq
+
+import (
+	"testing"
+
+	. "github.com/onsi/gomega"
+)
+
+func Test_ApplyFilter_SingleDocumentModification(t *testing.T) {
```
ldmonster, line 9: whats about handling empty filter?
i remember, that was a problem, when we have no result if filter is empty.

ldmonster (approved):

pkg/filter/jq/apply.go
```diff
@@ -0,0 +1,51 @@
...
+type Filter struct{}
+
+// ApplyFilter runs jq expression provided in jqFilter with jsonData as input.
+func (f *Filter) ApplyFilter(jqFilter string, data map[string]any) (map[string]any, error) {
+	query, err := gojq.Parse(jqFilter)
+	if err != nil {
+		return nil, err
+	}
+
+	iter := query.Run(data)
```
ldmonster, line 27: under hood this method [can change](https://github.com/itchyny/gojq/blob/main/normalize.go#L78) map from unput

juev: In order to ensure that the input data does not change, I created a separate data instance

ldmonster (approved):

yalosev (approved):

Issue #780, opened 2025-07-15
ghcr packages are stale! latest is v1.6.2 but v.19.0 in releases


vitalyu:
**Expected behavior (what you expected to happen)**:

Release version == Package version

PR #821, opened 2025-10-13
Fix release github workflow

verdel:
#### Overview

In PR #761, the event that triggers the GitHub workflow for publishing the Docker image was changed. However, the `check` job code was not updated accordingly. As a result, after the PR was merged, docker images stopped being published to the registry following the release of shell-operator.

I fixed the GitHub workflow so that when it is manually triggered from the main branch, the Docker image is published with the latest tag, and when a release is published, it is tagged with both the release tag and the latest tag.

This PR also fixes the issue #780.


ldmonster:
Fixed here https://github.com/flant/shell-operator/pull/832

PR #867, opened 2026-04-13
[chore] add typed error for executor

Fkuloff:
#### Overview
Replace fmt.Errorf("stderr: ...") with a custom StderrError type in RunAndLogLines. Callers can use errors.As to detect stderr errors and access raw content without brittle string prefix trimming. Error() returns the same string format — fully backward compatible.

### DIFF
--- a/pkg/executor/executor.go
+++ b/pkg/executor/executor.go
@@ -32,6 +32,16 @@ func Run(cmd *exec.Cmd) error {
 	return cmd.Run()
 }
 
+// StderrError is returned by RunAndLogLines when a command fails and produces
+// output on stderr. Callers can use errors.As to access the raw stderr content.
+type StderrError struct {
+	Stderr string
+}
+
+func (e *StderrError) Error() string {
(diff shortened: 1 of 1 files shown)


pkg/executor/executor.go
```diff
@@ -32,6 +32,16 @@ func Run(cmd *exec.Cmd) error {
 	return cmd.Run()
 }
 
+// StderrError is returned by RunAndLogLines when a command fails and produces
+// output on stderr. Callers can use errors.As to access the raw stderr content.
+type StderrError struct {
+	Stderr string
```
ipaqsa: Rename to `Message`

pkg/executor/executor.go
```diff
@@ -32,6 +32,16 @@ func Run(cmd *exec.Cmd) error {
...
+}
+
+func (e *StderrError) Error() string {
+	return fmt.Sprintf("stderr: %s", e.Stderr)
```
ipaqsa: Just `e.Message`

ipaqsa (approved):

PR #874, opened 2026-05-15
Bump golang version

n0npax:
#### Overview

some scanners are complaining about go/stdlib
which sec issues for go versions 0:1.23.8 / 1.26.0-rc.3

Bumping to `1.26.3`

keeps some sec scanners happy :)

golangci-lint was failing on `swag.YAMLToJSON` so switched to  `yamlutils.YAMLToJSON`.
Except that there are just version bumps


n0npax:
@ldmonster 🙏🏻

PR #876, opened 2026-05-16
fix(hook): keep counter add metrics ungrouped by default

### LINKED ISSUE #868: counter metrics do not count up
**Expected behavior (what you expected to happen)**:

using a counter metric, every time you add a value the counter should be increased by the value.

**Actual behavior (what actually happened)**:

the counter metric is set to the value it should be increased. it is not counting up.

**Steps to reproduce**:
1. have a hook with a counter metric and set "action":"add" and "value":1
2. execute this hook every minute
3. metrics is always set to 1 and not counting up

**Environment**:
- Shell-operator version: v1.14.3
- Kubernetes version: v1.34.4
- Installation type (kubectl apply, helm chart, etc.): long running container started from shell-operator docker image and hook is triggered  by cron config

**Anything else we should know?**:

This issue came up after updating from shell operator v1.6.2 to v1.14.3. with v1.6.2 the behavior was correct and the counter was counting up. Maybe this was introduced by https://github.com/flant/shell-operator/releases/tag/v1.11.0

aspr0d:
## What

Fixes #868

When a hook emits a counter with `action: add` and no explicit `group`, the parser assigned the hook name as the default group. Grouped metric batches expire all series in the group before applying new values, so counters were reset on every run.

## Changes

- Do not assign `defaultGroup` for `action: add` when `group` is omitted
- Gauges and histograms still get the hook name as the default group for expire-and-refresh semantics
- Tests cover parsing behavior and cumulative counter updates via `ApplyBatchOperations`

## Verification

`go test ./pkg/hook/ -run 'TestMetricOperations|TestCounterAdd' -count=1`

### DIFF
--- a/pkg/hook/operation.go
+++ b/pkg/hook/operation.go
@@ -80,7 +80,10 @@ func MetricOperationsFromReader(r io.Reader, defaultGroup string) ([]MetricOpera
 			metricOperation.Value = metricOperation.Add
 		}
 
-		if metricOperation.Group == "" {
+		if metricOperation.Group == "" && metricOperation.Action != "add" {
+			// Counter "add" operations without an explicit group must stay ungrouped so
+			// values accumulate across hook runs. Gauges and histograms still get the
+			// hook name as the default group for expire-and-refresh semantics.
 			metricOperation.Group = defaultGroup
 		}
 
--- a/pkg/hook/operation_group_test.go
+++ b/pkg/hook/operation_group_test.go
@@ -1,0 +1,107 @@
+package hook
+
+import (
+	"testing"
+
+	metricsstorage "github.com/deckhouse/deckhouse/pkg/metrics-storage"
+	"github.com/deckhouse/deckhouse/pkg/metrics-storage/operation"
+	. "github.com/onsi/gomega"
+)
+
+func TestMetricOperationsFromBytes_CounterAddWithoutDefaultGroup(t *testing.T) {
+	g := NewWithT(t)
+
+	ops, err := MetricOperationsFromBytes([]byte(`{"name":"my_counter","action":"add","value":1}`), "my-hook")
+	g.Expect(err).NotTo(HaveOccurred())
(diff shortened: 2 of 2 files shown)


ldmonster:
The correct way to use hook metrics with expire previous runs metrics.

aspr0d:
Thanks for the note. Could you clarify or share an example?

My understanding: counters with `action: add` are intended to accumulate across runs, so applying `ExpireGroupMetrics` before each batch resets them - which is the regression reported in #868.

If there is a preferred pattern (e.g. always requiring an explicit `group` for counters, or a separate expiry path for add vs set actions), I am happy to rework the fix to match it.

ldmonster:
> Thanks for the note. Could you clarify or share an example?
> 
> My understanding: counters with `action: add` are intended to accumulate across runs, so applying `ExpireGroupMetrics` before each batch resets them - which is the regression reported in #868.
> 
> If there is a preferred pattern (e.g. always requiring an explicit `group` for counters, or a separate expiry path for add vs set actions), I am happy to rework the fix to match it.

Thanks for the follow-up.

In the current design, accumulating metrics across hook runs is outside the scope of what hooks are meant to do. Shell-operator is a tool for running hooks in Kubernetes; hook metrics exist to track what happens inside a single hook execution, not to carry state from one run to the next.

The scenario described in #868 is effectively trying to use hook metrics as cross-run storage. That is not a supported use case, so resetting counters via ExpireGroupMetrics before each batch is consistent with the intended model rather than a regression.

If you need values to persist across runs, that should be handled outside the hook metrics layer—for example via Kubernetes objects, external storage, or another mechanism appropriate to your workflow—not by relying on action: add counters to accumulate between invocations.

Happy to discuss further if a concrete example would help clarify the boundary.

aspr0d:
Thanks for the detailed explanation - that makes the design intent clear.

Closing this PR. The behavior is correct as designed.

PR #886, opened 2026-05-22
[shell-operator] fix: track child PIDs to avoid zombie reaper race

fuldaxxx:
#### Overview
Adds a process registry to the `executor` package so that the PID-1 zombie reaper can distinguish between children it owns and children spawned by the executor (hooks, scripts). This prevents a race condition where `SIGCHLD` handling reaps a child that `cmd.Wait` still expects to collect, causing "signal: child exited already" errors and lost exit codes.

When `shell-operator` runs as PID 1 in a container, its zombie reaper loop calls `waitpid(-1, ...)` on every `SIGCHLD`. This can steal a child that was just started by the executor - `cmd.Start()` returns, the child exits, `SIGCHLD` fires, and the reaper reaps the process before `cmd.Wait()` gets to it. The result is a spurious `exec: signal: child exited already or missing exit-code propagation.`

 This PR introduces:

 - `processRegistry (registry.go)` — a thread-safe, package-internal map that tracks PIDs of processes started by the executor. Exposes a read-only `ProcessTracker` interface via `Tracker()` for external consumers (zombie reaper) and package-private `registerPID/unregisterPID` helpers for the executor methods.
 - Modified `Run`, `Output`, and `RunAndLogLines` in `executor.go` — split `cmd.Run() / cmd.Output()` into explicit `Start` → `registerPID` → `Wait` → `unregisterPID` sequences so the child PID is
   tracked for the entire lifetime between fork and reap.
 - `Comprehensive tests (executor_test.go)` — covers registry basics, double-unregister safety, concurrent access, and integration with `Output / RunAndLogLines` (including failed-start scenarios).

 The zombie reaper can now call `executor.Tracker().IsActive(pid)` to skip PIDs that the executor will reap itself.


ldmonster (approved):

pkg/executor/executor.go
```diff
@@ -24,12 +25,22 @@ const (
...
+// Run starts the command, waits for it to complete, and returns the error.
+// The child PID is registered in the global process registry while the process
+// is running so that a PID-1 zombie reaper does not steal it.
 func Run(cmd *exec.Cmd) error {
 	// TODO context: hook name, hook phase, hook binding
 	// TODO observability
 	log.Debug("Executing command", slog.String(pkg.LogKeyCommand, strings.Join(cmd.Args, " ")), slog.String(pkg.LogKeyDir, cmd.Dir))
 
-	return cmd.Run()
+	if err := cmd.Start(); err != nil {
```
ldmonster: Between line 36 (cmd.Start() succeeds → child is alive) and line 40 (registerPID), the PID exists in the kernel but is not in our registry. A fast-exiting child (e.g. echo, true) can exit() in those few microseconds and deliver SIGCHLD to the reaper. If the reaper's wait4(-1, WNOHANG) runs in that window and consults the tracker, it will see IsActive(pid) == false and reap the child — defeating the whole purpose of the registry. cmd.Wait() then returns ECHILD ("no child processes").

The same applies to Output() and RunAndLogLines().

pkg/executor/registry.go
```diff
@@ -0,0 +1,73 @@
...
+// processRegistry tracks PIDs of processes started by the executor so that
+// a PID-1 zombie reaper can skip them (their parent already calls Wait).
+// This prevents the reaper from stealing a child that cmd.Wait expects to reap.
+//
+// The struct is intentionally unexported — all external access goes through
+// the ProcessTracker interface (read-only) or the package-level helpers
+// registerPID / unregisterPID (write, executor-internal).
+type processRegistry struct {
+	mu         sync.RWMutex
+	activePIDs map[int32]struct{}
```
ldmonster: The registry stores PIDs as int32 but os.Process.Pid is int. The cast is unnecessary and risks truncation on extremely high pid_max values; also creates a minor API inconsistency.

fuldaxxx: done

pkg/executor/executor.go
fuldaxxx: done

